PR #44, opened 2020-04-18
Added Jack Liu to about page

jackliu612:


brian-yu:
@jackliu612 Changes should be live shortly on http://dev.thecourseforum.com/ !

PR #49, opened 2020-04-19
About jasmine

ejd5mm:
Added missing curly bracket and comma on line 110


brian-yu:
Changes are live on http://dev.thecourseforum.com/about !

Issue #12, opened 2020-04-10
Search


brian-yu:
Elasticsearch or similar integration for course/instructor/review search

Requirements
- overall website search (e.g. navbar)
- individual course / professor / semester search
  - e.g. search in the review creation form

bradknaysi:
done

PR #55, opened 2020-04-19
Elasticsearch index complete (for courses)

bradknaysi:
# What I did
- created a VERY robust `django-admin` command
- add instructions for using it
- loaded all the Course data to the production Elastic App Search cluster

Progress on the linked Issue #12 
- As you can see there are 14,579 courses, which you can cross-check with the actual database to verify the script worked. A nice thing about the script is that you can run it as many times as you want and it still works (b/c key=course.pk)

# Production Elastic App Search 'uva-courses' engine summary picture
uva-courses engine


tcf_website/management/commands/index_elasticsearch.py
```diff
@@ -0,0 +1,113 @@
...
+    many times as you want! It updates a document in Elastic if it always exists so won't
+    double add or anything weird like that!
+
+    """
+
+    help = 'Erases and re-indexes the Elastic hosted Elasticsearch cluster'
+
+    def handle(self, *args, **options):
+
+        courses_engine_endpoint = 'https://1761244fa3674318b11f8d5729ffd7a2.app-search.us-east4.gcp.elastic-cloud.com/api/as/v1/engines/uva-courses/documents'
```
brian-yu, line 40: You can move the endpoint and API key into the `.env` file and then access it like so:

`.env` file:
```bash
# .... current stuff above
ES_COURSE_ENDPOINT=https://1761244fa3674318b11f8d5729ffd7a2.app-search.us-east4.gcp.elastic-cloud.com/api/as/v1/engines/uva-courses/documents
ES_API_KEY=<key>
```

Then in `python` access them like this:
```python
import os

courses_engine_endpoint = os.environ['ES_COURSE_ENDPOINT']
api_key = os.environ['ES_API_KEY']
```

tcf_website/management/commands/index_elasticsearch.py
```diff
@@ -0,0 +1,113 @@
...
+                    # Convert list to json string
+                    json_documents = json.dumps(documents)
+
+                    # Send documents to the Elastic endpoint
+                    response = requests.post(
+                        url=courses_engine_endpoint,
+                        data=json_documents,
+                        headers=https_headers
+                    )
+                    self.stdout.write("Indexed courses: " + str(start) + " - " + str(end) + "   status_code=" + str(response.status_code))
```
brian-yu, line 85: Maybe this and lines 100-109 can be refactored into another function.

tcf_website/management/commands/index_elasticsearch.py
```diff
@@ -0,0 +1,113 @@
...
+                # Convert list to json string
+                json_documents = json.dumps(documents)
+
+                # Send documents to the Elastic endpoint
+                response = requests.post(
+                    url=courses_engine_endpoint,
+                    data=json_documents,
+                    headers=https_headers
+                )
+                self.stdout.write("Indexed courses: " + str(start) + " - " + str(start + count) + "   status_code=" + str(response.status_code))
```
brian-yu, line 109: Function?

tcf_website/management/commands/index_elasticsearch.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # PRIVATE READ / WRITE KEY (need to find more secret place)
+        api_key = 'private-tw8dxda1rfxw28s7dxc6vihs'
+
+        https_headers = {
+            "Content-Type" : "application/json",
+            "Authorization" : "Bearer " + api_key
+        }
+
+        all_courses = Course.objects.all().order_by('pk')
```
brian-yu, line 50: Should we re-index all courses or just courses last taught in the current semester?

tcf_website/management/commands/index_elasticsearch.py
bradknaysi: gotcha

tcf_website/management/commands/index_elasticsearch.py
bradknaysi: yeah probably should lol

tcf_website/management/commands/index_elasticsearch.py
bradknaysi: yeah probably should lol

tcf_website/management/commands/index_elasticsearch.py
bradknaysi: I say stick will all courses, because it will UPDATE if there are any changes and won't do anything if it's the same. This approach is more robust from future error and there will never be enough courses to make this an issue, even with 14000 courses it only takes ~120 seconds

brian-yu: Sounds good to me!

tcf_website/management/commands/index_elasticsearch.py
brian-yu: Can you add those to this document: https://docs.google.com/document/d/1aAV0D8uD84h-JxjP_j7a2yKZz8pFM36P2tltHj_G3m8/edit

bradknaysi: - added production environment variables ✅ 
- updated google doc ✅ 
- tested by changing my environment to production ✅

brian-yu: Can you split out those redundant lines into a function? After that I'll approve and merge

PR #81, opened 2020-04-23
App engine

saikonuri:
- [x] Set up uwsgi as the prod server
- [x] Create app.yaml template which will be populated in Travis build
- [x] Added Google App Engine encrypted service account key file
- [x] Added environment variables to Travis CI


app.yaml.template
```diff
@@ -0,0 +1,33 @@
...
+# The --http flag exposes the app with uWSGI only and will act as the router/proxy/load balancer
+# If we want an nginx webserver upstream for more features, we would have to use a --socket flag so that nginx can communicate with uWSGI
+# More info here: https://uwsgi-docs.readthedocs.io/en/latest/tutorials/Django_and_nginx.html
+entrypoint: uwsgi --http :8080 --module tcf_core.wsgi --master --processes 1 --threads 2
+
+handlers:
+# This configures Google App Engine to serve the files in the app's static
+# directory.
+- url: /static
+  static_dir: tcf_website/static/
```
brian-yu, line 14: do you know if we have to do `collectstatic` and then set `static_dir` to `STATIC_ROOT`? https://docs.djangoproject.com/en/3.0/howto/static-files/#deployment

tcf_core/settings.py
```diff
@@ -168,21 +169,26 @@
...
-        'thecourseforum.com',
-        'staging.thecourseforum.com',
-        'dev.thecourseforum.com',
-        'thecourseforum-staging.herokuapp.com',
-    ]
+    # SECURITY WARNING: App Engine's security features ensure that it is safe to
+    # have ALLOWED_HOSTS = ['*'] when the app is deployed. If you deploy a Django
+    # app not on App Engine, make sure to set an appropriate host here.
+    # See https://docs.djangoproject.com/en/1.10/ref/settings/ (from GCP documentation)
+    ALLOWED_HOSTS = ['*']
```
brian-yu, line 184: How will we point `thecourseforum.com` to the app engine instance? Would it be a `CNAME` DNS record to the app engine URL?

app.yaml.template
brian-yu: NVM I see that you changed `STATIC_ROOT`. do you think we need to run `collectstatic`?

tcf_core/settings.py
```diff
@@ -168,21 +169,26 @@
...
+
+    # ALLOWED_HOSTS = [
+    #     'thecourseforum.com',
+    #     'staging.thecourseforum.com',
+    #     'dev.thecourseforum.com',
+    #     'prod.thecourseforum.com',
+    # ]
```
brian-yu, line 191: Delete please

app.yaml.template
saikonuri: Yep you are right, I need to add `collectstatic`. Looking at the link I think there are a couple options we could use: 

- If we want to serve the static files from the same server, then we just run collectstatic before deploying and keep `STATIC_ROOT` as it is right now

- Or if we want a separate designated server for serving static files, we'll have to find a way to run `collectstatic` and push those files to that server. 

I personally think we don't need another server but what do you think?

tcf_core/settings.py
saikonuri: Perhaps something like this? 
https://cloud.google.com/appengine/docs/standard/python/mapping-custom-domains

app.yaml.template
brian-yu: I think we're good with 1 server for now. If needed, we can put them in a bucket or something later.

tcf_core/settings.py
saikonuri: So, what are we doing with the current heroku staging and dev instances?

tcf_core/settings.py
brian-yu: sweet! I can do that

app.yaml.template
saikonuri: Okay sounds good. I will add that `collectstatic`

tcf_core/settings.py
brian-yu: I believe that those define HOSTNAME, so that'll be handled in line `194`

saikonuri: Ah I see. Do I need to uncomment the django_heroku.settings(locals()) section to still be able to deploy to Heroku?

app.yaml.template
```diff
@@ -0,0 +1,33 @@
...
+
+# This handler routes all requests not caught above to your main app. It is
+# required when static routes are defined, but can be omitted (along with
+# the entire handlers section) when there are no static files defined.
+- url: /.*
```
brian-yu, line 19: should we add a `secure: always` to force https? https://stackoverflow.com/questions/5367974/https-only-in-google-app-engine

saikonuri: Added!

brian-yu:
Can you also do a `git restore --source=HEAD --staged --worktree -- tcf_website/static` in order to reset that directory?

https://stackoverflow.com/questions/15404535/how-to-git-reset-hard-a-subdirectory

Issue #75, opened 2020-04-20
Better Search


brian-yu:
See #79 

Improve upon the website's search.

Ask @brian-yu if you have any questions

Done - Basic search functions

In progress 
- Fix weird results (instructors)
- Improve UI

bradknaysi:
# Starting point... read the documentation!

[Elasticsearch Search API](https://swiftype.com/documentation/app-search/api/searchI)

bradknaysi:
# Update
### Todo
- improve instructor search algorithm
     - add weights
     - add boost for professors with recently taught courses

alex-shen1:
Too vague, if we revisit this at some point we'll make a new issue.

PR #88, opened 2020-04-28
Improved search algorithm (Course)

bradknaysi:
# What I did
Built a significantly better search algorithm for the `uva-courses` engine using the various query parameters that Elastic App Search API offers. I also added support for engine-specific search algorithms (e.g. course vs instructor).

This adds to the progress of #75


tcf_website/models/models.py
```diff
@@ -300,6 +300,10 @@ def average_difficulty(self):
         return Review.objects.filter(course=self).aggregate(
             models.Avg('difficulty'))['difficulty__avg']
 
+    def review_count(self):
```
brian-yu, line 303: I believe you could also do `self.review_set.count()`

tcf_website/views/search.py
```diff
@@ -110,3 +108,53 @@ def format_instructors(results):
         formatted.append(instructor)
 
     return formatted
+
+
+def build_elasticsearch_query(query, api_endpoint):
```
brian-yu, line 113: I think we should abstract away the api_endpoints. 

Maybe add `search_courses` and `search_instructors` that wrap `build_elasticsearch_query` and use os.environ directly so we don't have to keep passing around `api_endpoint`? Then lines 21 and 22 would be `search_courses(query)` and `search_instructors(query)`

tcf_website/models/models.py
bradknaysi: I wasn't gonna change it so the code stays consistent w/ the other course model methods

brian-yu: Hmm I'd argue the other methods (that I wrote haha) should be refactored to use `self.review_set.X` as it seems more readable

tcf_website/views/search.py
bradknaysi: done

tcf_website/models/models.py
bradknaysi: done

PR #90, opened 2020-04-30
Responsive google login button

bradknaysi:
# What I did
I made the login button dynamic so it responds to mouse hovering and clicks. I also updated the css to make the login page more responsive to scaling

new_login_button


brian-yu:
looks awesome!

PR #93, opened 2020-05-03
Add course sorting

jackliu612:
Added JS to sort classes in static/department as well as changes to the common/toolbar and department templates. Also updated .gitignore to ignore the static in the root directory only.


brian-yu:
Thanks, Jack! I'll give this a review.

brian-yu:
Also, in the future, please request a review from one of the exec so we know to review it!

tcf_website/static/department/sorting.js
```diff
@@ -0,0 +1,216 @@
...
+
+const sort_classes = (cmp) => {
+    $("ul.course-list").each(function (a) {
+        var li = $(this).children("li");
+        li.detach().sort(cmp);
+        $(this).append(li);
+    });
+}
+
+const cmp_by_rating_D = (a, b) => {
```
brian-yu, line 90: Thoughts on renaming to `cmp_by_rating_desc`? Might be a little more descriptive.

tcf_website/static/department/sorting.js
```diff
@@ -0,0 +1,216 @@
...
+    if (difficultyA > difficultyB) {
+        return 1;
+    } else if (difficultyB > difficultyA) {
+        return -1;
+    } else {
+        return 0;
+    }
+};
+
+const cmp_by_gpa_D = (a, b) => {
```
brian-yu, line 166: Does this handle when GPA is missing?

tcf_website/static/department/sorting.js
```diff
@@ -0,0 +1,216 @@
...
+        li.detach().sort(cmp);
+        $(this).append(li);
+    });
+}
+
+const cmp_by_rating_D = (a, b) => {
+    ratingA = a.querySelectorAll(".mb-0.info")[0].innerHTML.trim();
+    ratingB = b.querySelectorAll(".mb-0.info")[0].innerHTML.trim();
+
+    if (ratingA === '—') {
```
brian-yu, line 94: maybe `if (isNaN(ratingA)` so as to be more robust (in case we change the null signifier to some other character)

tcf_website/static/department/sorting.js
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    if (difficultyA > difficultyB) {
+        return 1;
+    } else if (difficultyB > difficultyA) {
+        return -1;
+    } else {
+        return 0;
+    }
+};
```
brian-yu, line 164: Is there a way to condense these into one function? Like have one function that can do both? It's OK if that's not easy to do-- then we'll just stick with having 2.

tcf_website/static/department/sorting.js
```diff
@@ -0,0 +1,216 @@
...
+    $("ul.course-list").each(function (a) {
+        var li = $(this).children("li");
+        li.detach().sort(cmp);
+        $(this).append(li);
+    });
+}
+
+const cmp_by_rating_D = (a, b) => {
+    ratingA = a.querySelectorAll(".mb-0.info")[0].innerHTML.trim();
+    ratingB = b.querySelectorAll(".mb-0.info")[0].innerHTML.trim();
```
brian-yu, line 92: Please make `const` declarations

brian-yu:
Also, how does this look when the breadcrumbs are super long?
e.g.
image

And can you attach a screenshot to the PR so that other people don't have to checkout locally and run?

Thanks! Again, awesome work. 😄

tcf_website/static/department/sorting.js
jackliu612: Not yet, I will update to include an `isNaN` check similar to the others

tcf_website/static/department/sorting.js
jackliu612: I was thinking about this and I wasn't sure if it was possible. It's not possible to simply reverse the array as I want the NaN elements to be at the bottom regardless of asc/desc. I'm also using the Array.sort() method which can take a comparator function but it needs to have two parameters only. I'm not sure if there's a way to implement a partial function of sorts to condense the asc/desc and maybe all the comparator functions as well.

jackliu612:
> Also, how does this look when the breadcrumbs are super long?

Unfortunately, it does spill over to a new line when the breadcrumbs get long such as here:
image
and it can get pretty cramped like here:
image

Not sure exactly how to resolve this.

tcf_website/static/department/sorting.js
brian-yu: Potentially (not sure if this is worth), you could filter out NaN elements, sort that, reverse if needed, and then concat with NaN elements. Might not be worth the additional complexity.

Another idea would be to 
```
const compare_by_difficulty_partial(ascending) {

    const compare_by_difficulty = (a, b) => {
        const difficultyA = a.querySelectorAll(".mb-0.info")[1].innerHTML.trim();
        const difficultyB = b.querySelectorAll(".mb-0.info")[1].innerHTML.trim();

        if (difficultyA === '—') {
            return 1;
        } else if (difficultyB === '—') {
            return -1;
        }

        if (difficultyA > difficultyB) {
            return Number(ascending);
        } else if (difficultyB > difficultyA) {
            return -1;
        } else {
            return Number(!ascending);
        }
    }

    return compare_by_difficulty;
}
```

brian-yu: Also, should we be using `parseFloat()` here?

brian-yu:
> Unfortunately, it does spill over to a new line when the breadcrumbs get long such as here:

How does it look if we put it underneath the semester selection button? Maybe we could do a `button-group-sm`? What if we merged the semester selection button group into your button group and made it one button that's a binary toggle? Just throwing ideas out there, not sure which is best.

brian-yu:
Another extreme nitpick: please use camel case (https://github.com/airbnb/javascript#naming--camelCase). We should add a JS linter that will fix things like this for us automatically.

tcf_website/static/department/sorting.js
jackliu612: I was simply doing string comparisons here as the ordering would be the same. I could convert to `parseFloat()` if that's better? Also, I think I might lean towards the partial implementation as I could then condense all of the functions into one.

jackliu612:
> How does it look if we put it underneath the semester selection button? 

I can try some of these options and see what looks best. I was originally thinking of a dropdown selection for sorting but I wasn't sure how to implement a toggle for order during that route.

tcf_website/static/department/sorting.js
brian-yu: You're right, string comparison is fine here. Maybe add a comment so that future readers won't think it's a bug?

brian-yu:
Thanks @jackliu612 for being so amenable to the code review comments! Just trying to make sure the code that we commit is maintainable and clean. That way y'all can uphold the standard after current exec graduates!

brian-yu:
@jackliu612 can this be closed because of #120?

PR #95, opened 2020-05-04
Twitter button now opens in a new tab

maralihart:


brian-yu:
Hi Mara, looks like there's no changes associated with this PR -- is there a commit that might still need to be pushed?

maralihart:
Oops, just pushed

PR #105, opened 2020-05-10
Django Messages support + "Already logged in" check

bradknaysi:
# What I did
I added banner style notifications to the application using the Django messages framework and Bootstrap 4 alerts. Now when you log out or add a review, it shows a temporary push notification. Additionally, I fixed a bug where you could be logged-in and still get to the login page.

# Picture (Final Design)
new tCF messages


brian-yu:
Awesome idea and work! I think that the message itself could look a bit better-- do you think you could get a design mockup from the design team?

bradknaysi:
I put in the UI mockup request!

brian-yu:
Brad, maybe you could use Bootstrap alerts? https://getbootstrap.com/docs/4.0/components/alerts/

j-alicia-long:
I agree with @brian-yu about keeping the flat material design of bootstrap alerts
I also recommend doing a lighter shade of orange so it's easier on the eyes

Additionally we might want to have different colors for success messages (blue?) and error messages (orange?)

image
^ This was what it looked like last I checked

bradknaysi:
I changed the messages UI to use bootstrap 4 alerts instead of my custom css. But, I found some backend issues with the default Django messages storage so I need to write a custom one. I'll re-request review @j-alicia-long and @brian-yu when I get it done. Might take a couple days.

https://docs.djangoproject.com/en/3.0/ref/contrib/messages/#storage-backends

bradknaysi:
I decided to stick with the default Django messages storage. It turns out that the problem was more-so something inherent with the way web-browsers operate. I decided to remove the problematic message, which was the "You're already logged-in" one.

This Pull Request is done and ready to be reviewed ✅

bradknaysi:
I didn't do automatic alert dismissal because it requires javascript and also the alert disappears following any page change or refresh anyway. This way, the implementation stays clean and robust. Another advantage is that the user won't miss seeing the alert!

PR #109, opened 2020-05-11
Review sorting

vn6:
Added review sorting


brian-yu:
Awesome work, Vi! I noticed there's some overlap in sorting functionality between @jackliu612 's recent PR (#93) and this. I wonder if we could have both of your PRs use some shared JS code to increase code reuse?

brian-yu:
@vn6 can this PR be closed?

Issue #99, opened 2020-05-09
Add Error handling for Elasticsearch failures


bradknaysi:
# Note
The error is because we haven't paid for our Elastic cluster yet and so they terminated it, so it's trying to reach the endpoint and finding nothing.

# Problem
The search code doesn't stably handle errors, resulting in the page crashing.

# Solution
Add error handling in `search.py` that display "search is down right now, try again later!" when something goes wrong. Make the notification for it look professional.

# Picture of the problem
<img width="1361" alt="tcf search server error bug" src="https://user-images.githubusercontent.com/30380036/81466269-58dd9b80-919e-11ea-9bab-8f69f9ddeb2f.png">

PR #110, opened 2020-05-11
ES error handling

bradknaysi:
# What I did
Added stable error handling to `search.py` so that it tells the user they can't search right now instead of crashing the page. I did a bunch of testing before reactivating the Elastic cluster so I know it works.

This closes issue #99


tcf_website/templates/search/search.html
```diff
@@ -14,6 +14,14 @@ <h1>Search: {{query}}</h1>
 
     <hr>
 
+    {% if courses_error %}
+    <div class="container">
+        <h4>Searching courses is down right now!</h4>
+        <li>{{courses_error}}</li>
+    </div>
+    {% endif %}
```
brian-yu, line 22: Should there be an `{% else %}` to display courses?

tcf_website/templates/search/search.html
```diff
@@ -14,6 +14,14 @@ <h1>Search: {{query}}</h1>
 
     <hr>
 
+    {% if courses_error %}
+    <div class="container">
+        <h4>Searching courses is down right now!</h4>
+        <li>{{courses_error}}</li>
```
brian-yu, line 20: Should we surface the error code to users?

tcf_website/templates/search/search.html
bradknaysi: true! that makes more sense

tcf_website/templates/search/search.html
bradknaysi: probably not, and not passing the error message will simplify the views code as well

tcf_website/views/search.py
```diff
@@ -103,6 +107,8 @@ def rank_course(query):
 
 def format_response(response):
     """Formats an Elastic search endpoint response."""
+    if "error" in response:
+        return response
```
jackliu612, line 111: This currently returns a dictionary, list, or string which seems awkward. Maybe change to always returning a list (empty if error) or returning two values, a list and a boolean/message/etc to signify the specific error raised? This would also make set_arguments simpler as we could do a check of `if courses:` instead of an isInstance check.

bradknaysi: Great suggestion! I went ahead and changed the `format_response` method to always return an error boolean and results list. Now the return type/format is standardized, which should fix all the edge cases. Go ahead and review the changes below and give the PR another review (or approval if you think its ready). Thank you for taking the initiative on this!

bradknaysi:
I disagree with this one. Although you want `success` in a response, you're left in the dark when there _wasn't_ `success`. By tracking `error`, we can answer the question of "why did this happen?" and thus offer the developer more understanding of the issue through its `message`. If there aren't any errors, then we're in the clear. If there were, then we have the information needed to debug it.

The user only cares about success, but the developer cares about error.

alex-shen1:
My bad — I overlooked `message` being a field that's created later because I was looking at where `formatted` was initially defined. I agree it should be `error`.

alex-shen1:
The branch is now out of date with `dev`, but other than that it looks good!

PR #122, opened 2020-05-16
Env security

j-alicia-long:
Moved secret keys from hardcoded in settings.py to .env file (and config vars on travis)


j-alicia-long:
Yes we're going to squash and merge LOL we finally got travis passing :')

j-alicia-long:
> Looks good to me! I'm concerned that everyone will not have the social auth, es private api key, etc. locally on their machines. How are we planning to have them set these variables on their machines?

We would put the local .env file in a shared access point, like on the google drive and link to it in the setup instructions for download

saikonuri:
> > Looks good to me! I'm concerned that everyone will not have the social auth, es private api key, etc. locally on their machines. How are we planning to have them set these variables on their machines?
> 
> We would put the local .env file in a shared access point, like on the google drive and link to it in the setup instructions for download

Sounds good!

bradknaysi:
@j-alicia-long make sure you merge this into `dev` not `master`. The `master` branch should never get ahead of `dev` commits-wise!

j-alicia-long:
@saikonuri can you approve review again since I change the base branch to dev

bradknaysi:
Looks like you need to fix merge conflicts, but other than that it looks good to me! Feel free to request review and I can approve.

j-alicia-long:
> In `search.py` and `index_elasticsearch.py`, environment variables are currently retrieved in the form `os.environ['...']`. Do these need to be changed to the new `env.str('...')` syntax?
> 
> If the answer to the above question is "no, those don't need to be changed" then this is ready to merge 👍

it's best if they are! I'll make the change

j-alicia-long:
actually it's probably fine I don't want to have to import and read the env file in multiple places

PR #137, opened 2020-06-05
Better "About Page"

bradknaysi:
# What I did
I re-did the About Page to fix a lot of formatting and css bugs. Now it looks much better, has a more accurate tCF history section, and uses Javascript to eliminate the page re-loading when switching between `history` versus `current team` versus `past contributors`. I also improved the about page file structure.


j-alicia-long:
Also i kinda miss the tagline under the jumbo logo

image

bradknaysi:
> One thing I will note is that the circular buttons (apply, write review) look kinda out of place between the rectangular blocks and tabs, but that can be fixed later on with another UI update.

I'm not gonna change them b/c I like the circular bubbles but if someone wants to change them later on that's fine.

PR #152, opened 2020-09-20
Fixing build issues

alex-shen1:


alex-shen1:
Added secrets as GitHub Secrets:

image

PR #153, opened 2020-09-20
Finish ESLint integration

Bartleby2718:
# Background
- [ESLint](https://eslint.org/) is a JavaScript linter, which we can use when we switch to [GitHub Super Linter](https://github.com/github/super-linter).

# Motivation
- A developer may forget to lint JavaScript files, so it makes sense to include linting during the CI process.

# What I Did
- Added `.eslintignore` so that some files are not linted by eslint
- Configured `.eslintrc.yml` to match our coding conventions
- Added `node_modules/` to `.gcloudignore` and `.gitignore` just in case
- Included an eslint command in `precommit` and `.travis.yml`
- Updated `Dockerfile` to install `nodejs` and `eslint`
- Updated `README.md` to include instructions on eslint (+a note on database migration)
- Added `package.json` and `package-lock.json`
- Ran `./precommit` to actually perform a lint

# Additional Notes
- Didn't work on Vagrant because Vagrant doesn't work on my machine and pylint isn't being run there as well
- I wasn't sure who I should add as reviewers/assignees or which labels to add, so I just kept them all empty.

# Screenshots
image
In this example, I intentionally took out the keyword `let` before the variable `sortNumberAsc`, and the statement that imports the function `cmpByProp`. and a statement that exports the function `sortByGpa`. These errors should be fixed manually by the developer, whereas other fixes like changing single quotes to double quotes or adding/removing whitespace characters are done automatically.


package.json
```diff
@@ -0,0 +1,27 @@
+{
```
j-alicia-long, line 1: Do you think we need npm to track js dependencies? I don't believe there's too many.
Just wanna avoid ballooning the repo with clutter

tcf_website/admin.py
```diff
@@ -51,7 +51,7 @@ class SectionAdmin(admin.ModelAdmin):
     autocomplete_fields = ['instructors']
 
     def get_queryset(self, request):
-        qs = super(SectionAdmin, self).get_queryset(request)
+        qs = super().get_queryset(request)
```
j-alicia-long, line 54: Not sure what these args were for...is this just a linting change? you sure we don't need those?

j-alicia-long (commented):
Good linting changes - overall looks good! 

Just have a few questions about the necessity of npm for this project - JS files comprise of only 1% of the codebase, so I'm not sure if it's worth the hassle (and added build time) to install everything in the image? Is it possible to run eslint without it?

tcf_website/admin.py
alex-shen1: Yeah pylint throws an error if it isn't set up like this

package.json
Bartleby2718: I'm a strong proponent of maintainability, and I believe being explicit is one way to achieve it. Also, I believe `npm install` gives you either errors or warnings if there is no `package.json`.

tcf_website/admin.py
Bartleby2718: @j-alicia-long  Yeah, there were some "errors" that were caught by pylint before I started working on ESLint. Seems like someone merged unlinted changes...

Bartleby2718:
> Just have a few questions about the necessity of npm for this project - JS files comprise of only 1% of the codebase, so I'm not sure if it's worth the hassle (and added build time) to install everything in the image? Is it possible to run eslint without it?

@j-alicia-long Well I assumed that the need for `ESLint` was agreed upon before I joined the team, so I looked up how to install ESLint and discovered that [it requires Node.js](https://eslint.org/docs/user-guide/getting-started#installation-and-usage). Also, you can choose to install `npm` and `eslint` in `precommit`, but `precommit` would be run more often than `docker build .`, so I think it'll take longer to build that way.

j-alicia-long:
> > Just have a few questions about the necessity of npm for this project - JS files comprise of only 1% of the codebase, so I'm not sure if it's worth the hassle (and added build time) to install everything in the image? Is it possible to run eslint without it?
> 
> @j-alicia-long Well I assumed that the need for `ESLint` was agreed upon before I joined the team, so I looked up how to install ESLint and discovered that [it requires Node.js](https://eslint.org/docs/user-guide/getting-started#installation-and-usage). Also, you can choose to install `npm` and `eslint` in `precommit`, but `precommit` would be run more often than `docker build .`, so I think it'll take longer to build that way.

Right, that makes sense! And @brian-yu added this task so I presume he knew what he's doing.

Wondering if we'll still npm once we integrate the super linter?

j-alicia-long (approved):
Looks great! Let's try out npm.

Issue #78, opened 2020-04-22
Performance profiling https://github.com/jazzband/django-silk


brian-yu:
Perhaps related to #11 

https://github.com/jazzband/django-silk

j-alicia-long:
Done (under score_display_performance branch)

Go to http://localhost:8000/silk/

Issue #11, opened 2020-04-10
Efficient review score aggregation


brian-yu:
Some way of caching or precomputing aggregate review scores so that we don't have to recompute on every page load.

https://math.stackexchange.com/questions/106700/incremental-averageing

Prefetching

j-alicia-long:
Proposal #1: Storing avg ratings in the Section model in avg_rating and update avg_rating on Review.save()
Would involve potentially creating new Section objects on review create

j-alicia-long:
**Issue 2:** Some department pages take longer to load than others (ex. Curry School, Architecture School)
**Solution Proposal 2:**  Listing smaller schools' departments on the browse page via dropdown, split into smaller views, so they aren't all loaded on the school-department page
- Would involve turning the smaller schools' subdepartments (ex. Kinesiology) into departments

PR #157, opened 2020-10-10
configured django-silk

j-alicia-long:
# What I did
Installed [django-silk](https://github.com/jazzband/django-silk) for performance profiling

## Issues addressed
#78 is resolved. Originally rolled it into a separate issue (#11 ) but decided to reopen since we're still addressing that.

# Screenshots
Created new url at http://localhost:8000/silk/. Can track load time for http requests and db queries as shown here: 
image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ tqdm>=4.45.0
 django-cachalot
 requests
 django-environ
+django-silk
 
 # Linting
 pylint>=2.4.4
--- a/tcf_core/settings.py
+++ b/tcf_core/settings.py
@@ -43,6 +43,7 @@ INSTALLED_APPS = [
     'django.contrib.staticfiles',
     'social_django',
     'cachalot',  # TODO: add Redis?
+    'silk', # Performance profiling
     'tcf_website'
 ]
 
@@ -54,6 +55,7 @@ MIDDLEWARE = [
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'silk.middleware.SilkyMiddleware', # Performance profiling
 ]
 
 ROOT_URLCONF = 'tcf_core.urls'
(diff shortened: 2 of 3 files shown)


Bartleby2718 (commented):
So are we allowing end users to access our performance data for all deployments (i.e. including the production website)? In terms of performance, installing one more package or running one more app isn't a big deal, but not sure if you'd want to share the data.

alex-shen1:
Yeah, seems like something we should remove at some point? Might be nice to have it stick around in dev branch, though; not sure if that makes it annoying to merge dev into master going forward because we have to specifically block one commit from going into prod each time we update master

alex-shen1 (commented):
As a side note, is there any particular rationale for the ordering of the apps/middlewares + if there isn't, is it worth having an ordering process?

Bartleby2718:
@alex-shen1 I was thinking of adding an `if` statement in `tcf_core/urls.py` so that the endpoint is not added in the production environment.

j-alicia-long:
> As a side note, is there any particular rationale for the ordering of the apps/middlewares + if there isn't, is it worth having an ordering process?

Yes, the ordering does matter to an extent, see below:
image
This is from the [django-silk docs](https://github.com/jazzband/django-silk) btw

j-alicia-long:
> @alex-shen1 I was thinking of adding an `if` statement in `tcf_core/urls.py` so that the endpoint is not added in the production environment.

Good idea! I will do that.

alex-shen1:
Re: keeping metrics private — if the tCF repo remains public, can't anyone just clone the repo and run it themselves to find out all this information? Pretty sure private repos cost money though so not sure what the solution is there

j-alicia-long:
> Re: keeping metrics private — if the tCF repo remains public, can't anyone just clone the repo and run it themselves to find out all this information? Pretty sure private repos cost money though so not sure what the solution is there

Actually, according to GitHub's current pricing plan (https://github.com/pricing) we now have unlimited private repos are for free accounts! Will make the change to private this week :)

alex-shen1 (approved):
Works on my machine too.

Bartleby2718 (approved):

PR #160, opened 2020-10-17
Django API with DRF

Bartleby2718:
# What I did
- Used [Django Rest Framework (DRF)](https://www.django-rest-framework.org/) to implement APIs for `School`, `Department`, `Subdepartment`, `Course`, `Instructor`, and `Semester` (I don't think we're using `Section`?)
- Paginated the `Course` API and the `Instructor` API
- Allowed filtering `Course` by `Subdepartment` ID and `Instructor` by `Couse` ID
- Made a minor change to `precommit` as it currently tries to auto-format other directories that are ignored by `git`

# Issues addressed
- You can now use JavaScript code to call the API and access the data needed to implement the dropdowns in the following mockup:
image
- Note that `Number` and `Instructor` need to be disabled initially and enabled only when the dropdown on their immediate left is selected (because there are countless `Number`s and `Instructor`s without any filter). Also, if you select something for `Prefix` and `Number` but change `Prefix` before selecting an option for `Instructor`, `Number` should be disabled again. See the following image from tCF1 production website:
image
- You may choose not to write JavaScript code for `Semester` as the server can fetch the small amount of data in an instant.

# Screenshots
- sample request & response:
image


Bartleby2718:
image
Actually I was able to achieve what is a prerequisite for faster loading of `/department/{id}` page: paginated `Course` data for each `Subdepartment` *with review statistics*

j-alicia-long:
Will give a comprehensive review in a bit - just wanted to note that Sections are created when semesters are updated (when we get new class registration info from the school)

It’s what links instructors to courses and semesters
Also I think we were going to rely on prepopulated sections to do “autocomplete” on review forms (ex. "Write a review" button on the course page)

ArjunK01 (commented):
looks good

tcf_website/api/serializers.py
```diff
@@ -0,0 +1,59 @@
...
+        fields = '__all__'
+
+
+class CourseWithStatsSerializer(serializers.ModelSerializer):
+    """DRF Serializer for Course including review statistics"""
+    semester_last_taught = SemesterSerializer(read_only=True)
+
+    class Meta:
+        model = Course
+        fields = ['id', 'title', 'description', 'number', 'subdepartment',
```
julesfll, line 50: Is it possible to include the course prefix for this endpoint?

julesfll (commented):
Looks good! A couple extra questions/comments:

- The `/?` in e.g. `/api/courses/?page=1` is a bit unconventional, could you remove the trailing slash? I saw [this](https://stackoverflow.com/questions/46163838/how-can-i-make-a-trailing-slash-optional-on-a-django-rest-framework-simplerouter) that might help
- It would be good to have the functionality to filter courses by semester (I might've missed it), that way users can sort by Fall 2020/All. This is probably possible in the frontend (not sure how it's implemented currently), but it would be good to have in the API
- It would be awesome if you could write up a simple doc for the API endpoints to share with the devs

tcf_website/api/serializers.py
Bartleby2718: @julesfll Because `subdepartment` is there, you can access it via `course.subdepartment.mnemonic`.

j-alicia-long (approved):
Merging this in to unblock the performance & review form PRs!

PR #163, opened 2020-10-18
Search style + department links

alex-shen1:
# What I did
Updated the search results page so that it's better than just a bulleted list. Results are sorted/grouped by department, with departments linked to. Probably not a particularly efficient way to implement the department link finding, but the primary bottleneck is ElasticSearch's API returning stuff anyways so it's not exactly the biggest problem IMO. 

One potential issue: as of this writing, this style takes up a whole lot more vertical space compared to the old bulleted list. Having users scroll is _probably_ not desirable, so maybe there's a way to compress it? 

## Issues Addressed
#75 kinda? I've interpreted that issue as generally "improve the search page," which this covers

# Screenshots

image


j-alicia-long:
I'll work on the UI for this!

j-alicia-long:
@alex-shen1 I did some work on styling w/ bootstrap: 

image

Only issue is that I'm unable to get the total course count for the "x results" text next to the Courses header. If you look in `search.py` you'll see some commented-out code where I attempted & failed to calculate the total course count from the query. If you could look at this that'd be great.

j-alicia-long:
Also @alex-shen1 are you still planning to work on the search filter options?

alex-shen1:
> Also @alex-shen1 are you still planning to work on the search filter options, as shown in Tina's mockup?

Not sure; at least in the mockup there are 2 fields and as far as I know we don't even have data for requirements yet, so only one filter is possible? Seems like a lot of work for one thing that isn't particularly helpful (especially considering the aforementioned 15 result cap; it really isn't that hard to read all 15 anyways)

j-alicia-long:
More updates on search page styling
image

alex-shen1:
Instructor page should really be its own branch....but it links to the search page so it's here now. Might branch off to flesh out instructors more later

j-alicia-long:
@alex-shen1 Make sure you're running `./precommit` to check tests & pylint before you push changes so your builds won't fail on Travis :)

alex-shen1:
whoops, should be fixed now!

j-alicia-long:
@alex-shen1 I get this bug when I go to http://localhost:8000/instructor/3291:
image

Since you just built the instructor pages, any idea what this is caused by?

alex-shen1:
Yup — after calculating average rating/difficulty I just rounded the value...forgot to check if the number actually exists at all. I'll have that fixed shortly!

vn6:
Looks good! One minor UI thing the "see results for" is not showing up for me
Capture

tcf_website/views/search.py
```diff
@@ -137,7 +140,7 @@ def format_courses(results):
         course = {
             "id": result.get("_meta").get("id"),
             "title": result.get("title").get("raw"),
-            "description": result.get("description").get("raw"),
+            # "description": result.get("description").get("raw"),
```
vn6, line 143: remove commented code

vn6 (approved):
looks good!

Issue #69, opened 2020-04-20
Send bug report via Discord bot (RIP Slack)


j-alicia-long:
include context about page, include user info

send message using discord API to dedicated channel

PR #164, opened 2020-10-18
Found a bug form

vn6:
# What I did
- Added a found a bug form that links to the sidebar
- Sends form results to discord (via webhook)
- Common form styling (linked to the bootstrap form-control class)
- Common form.js for validation (using [bootstrap validation](https://getbootstrap.com/docs/4.0/components/forms/#validation), checks fields with required tags)

## Issues addressed
#69 :)

# Screenshots
Click "Found a Bug?" on sidebar to open up form pop-up
form

Form validation w/ bootstrap will prevent submit if required fields are empty (this pic and the next one are from the old form, the pic above is more accurate but validation pretty much looks the same)
validation

Example of valid form inputs
valid

Confirmation message
confirm

Result is a message in discord #found-a-bug channel 
result


j-alicia-long (commented):
Looks good so far! I'd wait for Jihoon's jslint PR to be merged so you can run it on your js files.

tcf_website/static/base/bugform.js
```diff
@@ -0,0 +1,31 @@
...
+
+function postToDiscord(){
+  var url = $("#urlField").val();
+  var email = $("#emailField").val();
+  var description = $("#descriptionField").val();
+  var data = {
+    "content": "Bug Found! \n**URL:** " + url + "\n**Description**: \n" + description + "\n**Email:** " + email
+  }
+
+  var discordURL = "https://discordapp.com/api/webhooks/767189878223142942/wwBQA0K4VQ0i94ku2os2tYIVpPbDtfeP1i6s5G3CBWSCI7R0t6PbhZxwgJ8Z2yYpyv-q"
```
alex-shen1, line 22: Pretty sure we don't want this URL to be public? Maybe even regenerate a key (if that's possible? not sure how it works) if you can and then save it in `.env` (add it to the tCF secrets doc as well); it's highly unlikely that anybody would find it anyways, but I think it's good practice.

tcf_website/static/base/bugform.js
```diff
@@ -0,0 +1,31 @@
...
+  $.post( discordURL,
+          data,
+          function(){ console.log("posted to discord") });
```
alex-shen1, line 25: Log should be removed if we know the posting works now

tcf_website/templates/base/bugform.html
```diff
@@ -0,0 +1,50 @@
...
+        <h5 class="modal-title" id="bugModal">Report a Bug </h5>
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+        <form class="needs-validation" id="bugform" novalidate>
+          <div class="form-row">
+              <label for="urlField">URL</label>
+              <input type="text" class="form-control" id="urlField" placeholder="thecourseforum.com" value="thecourseforum.com" required>
```
alex-shen1, line 21: Should we do input validation? If you check the `found-a-bug` channel in the Discord you can see where I've made fake bug reports with bogus URLs and emails; not sure how easy it'd be to verify URLs, but emails should be relatively more straightforward

tcf_website/templates/base/bugform.html
```diff
@@ -0,0 +1,50 @@
...
+              <div class="valid-feedback">
+                Looks good!
+              </div>
```
alex-shen1, line 24: Where do these divs (the one here and the one for the description field) show up in the form? I can't figure out where to find them

alex-shen1 (changes requested):
Nice job! Just a few small comments, but overall having this working is great.

tcf_website/static/base/bugform.js
vn6: yeah probably not, for the .env to work here we would need to `npm install dotenv`, so I was trying to avoid that initially, should be a quick fix though

tcf_website/templates/base/bugform.html
vn6: Right now the validation only checks if there is any content. For the URLS, I could just set the field so it can't be modified unless you could think of a case when someone might need to change the URL?

tcf_website/static/base/bugform.js
alex-shen1: @j-alicia-long Looks like another push towards including npm in the repo...

tcf_website/templates/base/bugform.html
vn6: They only show up after the form has been validated and has not been submitted yet (see second screenshot). If everything is good, the form will just submit and close itself.

alex-shen1: Gotcha, cool

tcf_website/templates/base/bugform.html
alex-shen1: Honestly, we might be better off making that field optional, or entirely hiding the URL field from the form and just saving it behind the scenes — not sure how many users would either 1) go out of their way to copy a URL just to put in the form or 2) happen to exactly be on the page with the error when they click the bug form.

Another thing to consider is a case where the page crashes when accessing a certain URL, in which case it'd obviously be impossible to access the bug report form from that URL.

alex-shen1: Maybe create categories like "Crash," "UI/styling error," etc. to help point us in the right direction when looking @ bugs, might be easier than asking them to explicitly give the URL?

tcf_website/static/base/bugform.js
vn6: done

tcf_website/templates/base/bugform.html
vn6: I just took off the url field but am still sending that info to discord. I think the description part should probably cover all the categories, it might get too busy if we have too many options.

alex-shen1: yeah, the main idea was just to help us deal w/ reports more efficiently with some categorization, but now that I think about it there probably won't exactly be an overwhelming number of people who go out of their way to do bug reports.

j-alicia-long:
@vn6 Make sure you're running `./precommit` before you push changes so your builds won't fail on Travis :)

tcf_website/templates/base/bugform.html
j-alicia-long: I like @alex-shen1's idea of having bug report categories! Users aren't always as descriptive as we'd like them to be.

Having three inputs for "email", "category", and "description" isn't tooo busy (same number as before)

vn6: Makes sense, I can def add that in. Just thinking that we should limit the number of categories to choose from since if we have one for each feature that might be overwhelming. 

For now will be adding: Crash, UI/Styling, Data Error, Account Error, Other.  @alex-shen1 @j-alicia-long lmk if you have any other ideas/comments for categories.

vn6: also there's a few options for how to display the categories options
1. multiselect 
multiselect
2. select - can only choose 1 option tho
select
3. checkboxes - most similar to write a review mockup, can do horizontal (like below, if everything fits on one line) or vertical
check

vn6:
> @vn6 Make sure you're running `./precommit` to check tests & pylint before you push changes so your builds won't fail on Travis :)

will do, looks like eslint got me that time :,)

tcf_website/templates/base/bugform.html
vn6: done :)

tcf_website/static/base/bugform.js
vn6: Turns out you can only access the `.env` on the backend so no npm needed. Time to learn Django REST tho

alex-shen1:
@vn6 Can you add `DISCORD_URL_BUG` to the .env in [the tCF drive](https://docs.google.com/document/d/1HsuJOf-5oZljQK_k02CQhFbqw1q-pD_1-mExvyC1TV0/edit)?

tcf_website/static/base/bugform.js
vn6: > @vn6 Can you add `DISCORD_URL_BUG` to the .env in [the tCF drive](https://docs.google.com/document/d/1HsuJOf-5oZljQK_k02CQhFbqw1q-pD_1-mExvyC1TV0/edit)?

will do after I get it working :)

vn6: done

j-alicia-long (dismissed):
Looks good!

Dockerfile
```diff
@@ -1,18 +1,23 @@
-FROM python:3.6.5
+FROM ubuntu:bionic
+
+# ENV PYTHONUNBUFFERED 1
+
```
alex-shen1, line 4: What's going on here? As far as I know, we switched off `ubuntu:bionic` to `python:3.6.5` to improve build times... why is it changing back?

README.md
```diff
@@ -57,15 +57,20 @@ If you are part of theCourseForum engineering team, follow the instructions belo
         cd theCourseForum2
     ```
 3. Copy the [project secret keys](https://docs.google.com/document/d/1HsuJOf-5oZljQK_k02CQhFbqw1q-pD_1-mExvyC1TV0/edit?usp=sharing) into a `.env` file in the project base directory `theCourseForum2/`
-4. In the `theCourseForum2/` directory, run `docker-compose up` to start your Docker containers (tCF Django app and database)
+4. In the `theCourseForum2/` directory, run these commands to start your Docker container:
+    ```
+        docker build .
+        docker-compose up
+    ```
```
alex-shen1, line 64: Same thing here...? This is reverting stuff that shouldn't be reverted.

tcf_core/settings.py
```diff
@@ -245,15 +251,4 @@
...
-    """Exclude API views for django-silk"""
-    return not request.path.startswith('/api')
-
-
-# Performance profiling for non-API views during development
-SILKY_INTERCEPT_FUNC = custom_recording_logic
-if DEBUG:
-    INSTALLED_APPS.append('silk')
-    MIDDLEWARE.append('silk.middleware.SilkyMiddleware')
-    SILKY_INTERCEPT_FUNC = custom_recording_logic
```
alex-shen1, line 259: This should definitely stay here

alex-shen1 (changes requested):
I don't know why, but this entire PR is reverting a bunch of files that I'd assume have nothing to do with the review form....don't know why, but definitely can't merge in this state.

alex-shen1:
There's a _lot_ more files that it's affected, I just stopped reviewing once it became clear what was going on. I tried `git merge dev`, but it claimed that everything was up to date.....dunno what could've happened there.

alex-shen1:
Maybe cherry-pick commits out of the files that were meant to be changed and dump all the rest? Not quite sure on how to do that, though.

alex-shen1 (approved):

PR #165, opened 2020-10-18
Grades ui

ArjunK01:
# What I did
Just showing progress on grades ui 
Added chart.js and did some bootstrap formatting

## Issues addressed

# Screenshots
Screen Shot 2020-10-18 at 12 09 30 AM


j-alicia-long (commented):
@ArjunK01 @alex-shen1 Y'all consulted and decided to import the Chart.js bundle in order to avoid using npm, right? 

We should consult someone to see if it's better to use npm or do this...Jihoon's PR (#153 ) is currently blocked for the same reason.
@brian-yu @bradknaysi @saikonuri ?

.env 2.travis
```diff
@@ -0,0 +1,10 @@
+# Dev settings
```
Bartleby2718, line 1: Is this file supposed to be committed?

LICENSE 2
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
Bartleby2718, line 1: Is this file supposed to be committed?

pull_request_template 2.md
```diff
@@ -0,0 +1,5 @@
+# What I did
```
Bartleby2718, line 1: Is this file supposed to be committed?

tcf_website/static/about/history/history 2.css
```diff
@@ -0,0 +1,115 @@
+/*
```
Bartleby2718, line 1: Also looks like a duplicate

Bartleby2718 (changes requested):
There are so many duplicate files that it is hard to identify which changes are actual changes. Please let me know when you're done deleting them.

Issue #16, opened 2020-04-10
Grades data


brian-yu:
Done
- Backend (google sheets scraping)
- Grade calculations

In progress 
- Resolving build issues
- UI integration

bradknaysi:
Any updates on this?

PR #175, opened 2020-10-25
Grades Data

jackliu612:
# What I did
Added a script under management/commands to load grades data into the database
Created two new models CourseGrade and CourseInstructorGrade to store the grades data
Added xlrd to requirements 
Code mostly written by Neal and Harshita last year

## Issues addressed

#16 

# Screenshots
image
image


Bartleby2718 (commented):
When exactly are these scripts run? I've been thinking packages like `pandas` or `xlrd` can be moved to another file (e.g. `requirements.dev.txt`) so that they are not installed when not needed.

tcf_website/management/commands/load_grades.py
```diff
@@ -0,0 +1,398 @@
...
+            action='store_true',
+            help='Verbose output',
+        )
+
+        parser.add_argument(
+            'semester',
+            help='Semester to update grades (e.g. "2019_FALL").\nIf you wish to reload all semesters (potentially dangerous!) then put "ALL_DANGEROUS" as the value of this argument.',
+            type=str
+        )
+    '''
```
alex-shen1, line 41: Why is this a docstring? If it's just old code, then it might be good to just remove it?

tcf_website/management/commands/load_grades.py
```diff
@@ -0,0 +1,398 @@
...
+        [self.load_semester_file(file) for file in sorted(
+            os.listdir(self.data_dir)) if not file.startswith('.')]
+
+        '''
+        if semester == 'ALL_DANGEROUS':
+            # ignores temporary files
+            [self.load_semester_file(file) for file in sorted(os.listdir(self.data_dir)) if not file.startswith('.')]
+        else:
+            self.load_semester_file(f"{semester.lower()}.xlsx")
+        '''
```
alex-shen1, line 60: Same thing about cleanup here

tcf_website/management/commands/load_grades.py
```diff
@@ -0,0 +1,398 @@
...
+                'B+',
+                'B',
+                'B-',
+                'C+',
+                'C',
+                'C-',
+                'D+',
+                'D',
+                'D-',
+                'F'])
```
alex-shen1, line 81: Seems like this could be compressed a few lines...

tcf_website/management/commands/load_grades.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def load_semester_file(self, file):
+        year, semester = file.split('.')[0].split('_')
+        year = int(year)
+        season = semester.upper()
+
+        print(year, season)
```
alex-shen1, line 88: Should prints like this only be shown if verbose mode is on?

tcf_website/management/commands/load_grades.py
```diff
@@ -0,0 +1,398 @@
...
+                c_plus,
+                c,
+                c_minus,
+                d_plus,
+                d,
+                d_minus,
+                f,
+                ot,
+                drop,
+                withdraw]
```
alex-shen1, line 158: Same thing about compression here

tcf_website/management/commands/load_grades.py
```diff
@@ -0,0 +1,398 @@
...
+
+            # load this semester into course dictionary
+            if course_identifier not in course_grades:
+                course_grades[course_identifier] = this_semesters_grades
+            else:
+                for i in range(len(course_grades[course_identifier])):
+                    course_grades[course_identifier][i] += this_semesters_grades[i]
```
alex-shen1, line 165: General concern is that this overlaps data for Special Topics (e.g. CS 1501 or whatever) courses...but I can't think of a meaningful way to fix this.

tcf_website/management/commands/load_grades.py
```diff
@@ -0,0 +1,398 @@
...
+            if course_instructor_identifier not in course_instructor_grades:
+                course_instructor_grades[course_instructor_identifier] = this_semesters_grades
+            else:
+                for i in range(
+                        len(course_instructor_grades[course_instructor_identifier])):
+                    course_instructor_grades[course_instructor_identifier][i] += this_semesters_grades[i]
+
+        except TypeError as e:
+            print(row)
+            print(e)
```
alex-shen1, line 177: Same thing about verbose mode

tcf_website/management/commands/load_grades.py
```diff
@@ -0,0 +1,398 @@
...
+            2.3,
+            2.0,
+            1.7,
+            1.3,
+            1.0,
+            0.7,
+            0.0,
+            0.0,
+            0.0,
+            0.0]
```
alex-shen1, line 198: Same thing about compression

tcf_website/management/commands/load_grades.py
```diff
@@ -0,0 +1,398 @@
...
+            d_plus = int(row['D+'])
+            d = int(row['D'])
+            d_minus = int(row['D-'])
+            f = int(row['F'])
+            ot = int(row['OT'])
+            drop = int(row['DR'])
+            withdraw = int(row['W'])
+            # credit
+            # general_credit
+            # no_credit
```
alex-shen1, line 132: Seems like these 3 fields were originally planned to be part of the model, but they decided against it — should remove these comments at some point

alex-shen1 (commented):
Most of my points are just for the interest of code cleanup, but (granted I don't know much about data parsing) the functionality seems to be sound? @jackliu612 as I understand it this isn't your code, so you shouldn't necessarily be responsible for fixing it, so I can take it if you're busy. One thing is that I'm not sure off the top of my head how to do set up verbose logging, and seeing how you've done that for the other command it might be easier if you take that.

Other stuff like removing comments, refactoring etc. I can handle.

tcf_website/management/commands/load_grades.py
Bartleby2718: This seems to be code copied and adapted from `load_semester.py`. I think it's there because the functionality is not complete.

tcf_website/management/commands/load_grades.py
Bartleby2718: This is just a matter of preference. I will make the changes, but to play devil's advocate, it's easier to count the number of total lines, the index of each line, and every addition/deletion translates to addition/deletion of a single line (which looks better in `git diff`s).

tcf_website/management/commands/load_grades.py
Bartleby2718: Removed in c45f4454ed25cb07eb138184b1b935b9661baefb

tcf_website/management/commands/load_grades.py
Bartleby2718: Done in 2b434da273891c45744aeecfd86df860dbb10112

tcf_website/management/commands/load_grades.py
Bartleby2718: Done in 2b434da273891c45744aeecfd86df860dbb10112

tcf_website/management/commands/load_grades.py
Bartleby2718: I don't think anyone can fix this because this isn't something SIS considered in the beginning. Sherriff and McBurney both teach CS 3240 this semester, and their classes are considered a single class. That is, students are in different sections but listen to the same lectures and are graded on the same scale. On the other hand, Shen and Magee both teach CS 4740 this semester, but the only thing shared between the two classes are the name of the class. They are taught different materials, have different assignments, and graded on different scales. Ideally, we need another identifier for each section, and which should allow us to group different sections based on that, but we don't have that data.

tcf_website/management/commands/load_grades.py
Bartleby2718: Done in b11e94fd3d58db1a73892c61ad5720091ceecc7b

Bartleby2718:
Clsoing without merging for now

PR #182, opened 2020-10-26
Performance improvement for department pages

### LINKED ISSUE #171: Performance improvement for department pages
Infinite scroll pagination on class page list

Bartleby2718:
# Status
- Done

# What I did
- Used pagination and client-side rendering to dramatically improve the performance of /department/{id} page, which was extremely slow because the server rendered the entire page

## Issues addressed
- This should fix #171.

## Screenshots
- This should really be checked by yourself or through a video, but here's a screenshot:
image
- The courses will be rendered as responses start coming back, and the end result will be the same. However, the user can see *some* courses earlier.

## Note
- The worst case scenario is I believe `/department/{68}`, Curry School of Education.
- I have considered using server-side rendering for "the first page" and using client-side rendering for the remaining pages, but because "the first page" is the first page out of all courses offered in the last 5 years, so the user won't see any differences, unless the class is very new.


tcf_website/templates/base/base.html
```diff
@@ -46,9 +46,9 @@
 
     </div>
     <!-- /#wrapper -->
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js" integrity="sha384-ZvpUoO/+PpLXR1lu4jmpXWu80pZlYUAfxl5NsBMWOEPSjUn/6Z/hRTt8+pR6L4N2" crossorigin="anonymous"></script>
```
j-alicia-long, line 49: Thanks for changing the JS/JQuery to CDN as well!

tcf_website/templates/department/department.html
```diff
@@ -18,14 +18,13 @@
...
-                {% for c in sd.recent_courses|dictsort:"number" %}
-                    <li class="{% if not c.is_recent %}old{% endif %}">
-                        {% url 'course' course_id=c.pk as link %}
-                        {% include "common/rating_card.html" with title=c.code subtitle=c.title link=link rating=c.average_rating difficulty=c.average_difficulty last_taught=c.semester_last_taught content=c.description prefix='course' pk=c.pk %}
-                    </li>
-                {% endfor %}
+            <ul class="course-list" id="courses-sd-{{sd.id}}">
             </ul>
+            <div class="text-center" id="spinner-sd-{{sd.id}}">
+                <div class="spinner-border" role="status" >
```
j-alicia-long, line 24: Hmm, wonder if it'll look better if the spinner was in a white card instead of being directly on the background of the page

tcf_website/static/department/loadMoreCourses.js
```diff
@@ -0,0 +1,86 @@
...
+        loadPage(subdepartmentId, courses.next);
+    }
+}
+
+function generateCourseCardHTML(course) {
+    const rating = emdashOrTwoDecimals(course.average_rating);
+    const difficulty = emdashOrTwoDecimals(course.average_difficulty);
+    const seasonAllCaps = course.semester_last_taught.season;
+    const seasonTitleCase = seasonAllCaps.charAt(0) + seasonAllCaps.substr(1).toLowerCase();
+    return `
```
j-alicia-long, line 26: Is it possible to move this HTML to a template and just pass in param values you need? Would prefer to keep presentation separate from functionality.

j-alicia-long (changes requested):
Looks good, just left a few comments about code cleanliness/styling. Going to wait for a second reviewer as well.

By the way, is your API annotation documentation going to be in a separate PR?

tcf_website/templates/department/department.html
Bartleby2718: That's probably better. Ideally, it's best to have a placeholder, and then replace it with either the actual data or a box indicating the empty state. An example of placeholders from Instagram:
image
Since the main purpose of this PR is not design improvement, we can do that in another PR.

tcf_website/static/department/loadMoreCourses.js
Bartleby2718: I've thought about it and did some research, but I can't think of a way to do that. I'll ask someone with more experience.

Bartleby2718:
> By the way, is your API annotation documentation going to be in a separate PR?

@j-alicia-long  Yes, I'm thinking of updating `README.md` later, along with some other changes.

Bartleby2718:
Benchmarks on `/department/68`:
- After 6c21fcb1ba245ab4c8c972ad5ec521f0f775e082, now it takes 22-25 seconds to load and render everything on my machine (down from 40-44 seconds at ae2fee9e38e30eb131378bdc626a2dc1efef5376)
- It takes 22-24 seconds to load and render everything on https://thecourseforum-dev.herokuapp.com (whose HEAD is now at 03f1f12a1799703ba9884c172cf264b562ca2397 or, in terms of this PR, ae2fee9e38e30eb131378bdc626a2dc1efef5376 at the time of writing).
- Considering the difference on both machines, I expect the last few commits to give us roughly 2x improvement.

tcf_website/models/models.py
```diff
@@ -290,16 +290,6 @@ def is_recent(self):
...
-    def average_rating(self):
-        """Compute average rating."""
-        return Review.objects.filter(course=self).aggregate(
-            models.Avg('recommendability'))['recommendability__avg']
-
-    def average_difficulty(self):
-        """Compute average difficulty score."""
-        return Review.objects.filter(course=self).aggregate(
-            models.Avg('difficulty'))['difficulty__avg']
-
```
alex-shen1, line 302: Are you sure we should remove these? I assume they're meant to be replaced by API functionality, but I imagine there's some possibility that we may need them again for something in the future, and it might be easier/convenient to develop using this normal Django way if it's possible. If the API is strictly better/faster, then it makes more sense.

alex-shen1 (commented):
Looks good! I can't reproduce it, but I think at one point there was a bug where the course items (in `<li>` tags) showed up with bullet points on the left side. Not really sure why/how that happened, and it hasn't happened since; any idea? (It's entirely possible that I might've been in the wrong checked-out branch, or the files hadn't updated in time after switching branches, so I don't think this is a big issue)

Also, this branch is pretty out of date with `dev` so the styling doesn't match up, but I'm not quite sure how to resolve the conflict in `base.html` properly (the other conflict is trivial; just accept your deletion of the bootstrap file) so I'll leave it to you.

tcf_website/api/views.py
```diff
@@ -30,10 +31,25 @@ class SubdepartmentViewSet(viewsets.ReadOnlyModelViewSet):
 
 class CourseViewSet(viewsets.ReadOnlyModelViewSet):
     """DRF ViewSet for Course"""
-    queryset = Course.objects.all().order_by('number')
+    queryset = Course.objects\
+        .select_related('subdepartment', 'semester_last_taught')
     pagination_class = MyPagination
```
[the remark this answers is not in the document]
Bartleby2718: @alex-shen1 Any suggestion? I was thinking we need a global prefix (e.g. `TCF`) we can use for python/css/javascript classes or variables like this one.

tcf_website/models/models.py
Bartleby2718: Done in 1d04b3e2c44cb0af4b7ee98cb41da7793d7bcfb9

tcf_website/api/views.py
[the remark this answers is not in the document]
Bartleby2718: Renamed `MyPagination` to `FlexiblePagination` for now in e1e8331513eed2217df87ceead346429e57ad533
Let me know if you have a better name

Bartleby2718:
@alex-shen1 I think it's a temporary issue caused by browser cache not being properly invalidated, but let me know if that happens again. 

Did take care of the merge conflicts but that somehow made this PR merged...

tcf_website/static/department/loadMoreCourses.js
Bartleby2718: Didn't get a helpful response, and I can't think of a better way. I guess we can refactor this later?

tcf_website/templates/department/department.html
Bartleby2718: @j-alicia-long We can track this in https://github.com/thecourseforum/theCourseForum2/issues/191.

Issue #184, opened 2020-10-27
Slim down Dockerfile


Bartleby2718:
`docker build` takes too long, possibly because I'm out of the States. I just built `Dockerfile` from scratch (i.e. after removing the local `ubuntu:bionic` image), and it took 13 minutes on my Windows machine:
```
cmd /v:on /c "echo !time! && docker build . > before.txt && echo !time!"
 0:38:50.88
 0:51:56.80
```
Replacing with it with `python` (which includes `python3` and other basic packages like `curl`) shortened the build time significantly:
```
cmd /v:on /c "echo !time! && docker build . > after.txt && echo !time!"
 0:56:59.03
 1:00:12.82
```

PR #185, opened 2020-10-27
Slim down Dockerfile

Bartleby2718:
# Status
Done

# What I did
Changed the base image from `ubuntu:bionic` to `python:3.6.5`, requiring fewer packages to download, thus fewer network roundtrips and less network bandwidth

## Issues addressed
#184

## Screenshots
`cmd /v:on /c "echo !time! && docker build . > before.txt && echo !time!"` took 13 minutes on my Windows machine with the following output:
```
0:38:50.88
0:51:56.80
```
part of `before.txt`:
```
1 upgraded, 111 newly installed, 0 to remove and 0 not upgraded.
Need to get 115 MB of archives.
After this operation, 374 MB of additional disk space will be used.
```

---

After the change, `cmd /v:on /c "echo !time! && docker build . > after.txt && echo !time!"` took 3 minutes on my Windows machine with the following output:
```
0:56:59.03
1:00:12.82
```
part of `before.txt`:
```
4 upgraded, 8 newly installed, 0 to remove and 126 not upgraded.
Need to get 23.8 MB of archives.
After this operation, 99.2 MB of additional disk space will be used.
```


alex-shen1 (approved):
Works on my machine — nice work!

Issue #174, opened 2020-10-24
Fix review sorting by rating value


j-alicia-long:
Reviews on course page are currently sorted by absolute value (doesn't work for reviews with negative votes)

PR #186, opened 2020-10-28
Review Sorting Bug Fix

vn6:
# Status
Done? but cant really test it :(

# What I did
- updated the regex that extracts numbers from HTML to include negatives and decimals
- updated review page so that the sorting dropdown only appears if there are reviews

## Issues addressed
#174


j-alicia-long (commented):
Code looks good! If you can't test it locally how about writing a few django tests? Upvote is in models so you should be able to call it.

j-alicia-long (dismissed):

j-alicia-long (approved):

PR #196, opened 2020-11-08
Added some legacy fields to Review model

alex-shen1:
# Status
Done, In progress, etc

# What I did

tCF 1 keeps track of "fun" (`enjoyability` in the schema) ratings. We were missing this, but it was in legacy data so I updated our migration process/models to include this.

Also changed unit tests to account for the fact that calculating a Course's `rating` now includes enjoyability (which was obviously previously removed).

NOTE: will require redoing a bunch of legacy data migration stuff; @jackliu612 and I are in the middle of creating a new SQL dump file for devs.


alex-shen1:
Also branched off #162 so the number of commits being merged will drop after that one is merged

alex-shen1:
# Status
Backend done, perhaps frontend needs some more polishing

# What I did
Redoing this because the scope of the PR has changed dramatically lol

Added `enjoyability`, `amount_reading`, `amount_writing`, `amount_group`, and `amount_homework` fields to Review model, which were fields that existed in the legacy data. Created new data dump file (`updated_models_dump.sql` in the tCF drive)which should replace the old `april7.sql` we've been using for dev.

TODO: `hours_per_week` is still a field in the model even though it just sums up the 4 categories...could replace that with a function in the model that computes it instead of storing it straight up...but then I'd have to do a redump and I don't really want to do that. I can, but it's a lower priority thing.

legacy_db/mysql2sqlite2
```diff
@@ -0,0 +1,289 @@
+#!/usr/bin/awk -f
+
```
j-alicia-long, line 2: Add documentation to top of document saying what this is for (since you didn't write it)

reset_db
```diff
@@ -0,0 +1,10 @@
+sudo docker exec -i tcf_db psql -U tcf_django tcf_db
```
j-alicia-long, line 1: Flagged for deletion

tcf_core/settings.py
```diff
@@ -135,10 +135,14 @@
         'PASSWORD': 's3kr1t',
         'HOST': 'tcf_db',
     },
+    # 'legacy': {
```
j-alicia-long, line 138: Remove comments

j-alicia-long (approved):

PR #211, opened 2020-11-13
Vote fix

julesfll:
# Status
Done, In progress, etc

# What I did
- Reverted `review.js` to a regular import instead of a module import
 - Added `.eslintignore` to prevent eslint from complaining about that
- Reverted onclick tag for upvotes and downvotes
- Removed event listener script
- Removed `review.js` import from `course_professor` (unrelated, but it's already imported by `reviews.html`

## Issues addressed
- Upvotes and downvotes not working


j-alicia-long (approved):
Looks good, thanks for getting this working again! 

When we have more time we can look into fixing the event listener method to comply with eslint.

Issue #189, opened 2020-11-01
Review voting bug


j-alicia-long:
Review voting is broken on dev currently. Need to assess

Bartleby2718:
Resolved in #217

PR #214, opened 2020-11-14
fixed sorting by text comparison instead of numerical

jackliu612:
# Status
Done, In progress, etc

# What I did

Changed the common sorting script to sort reviews/classes using numerical comparison rather than string comparison (35 > 2). Very simple two-line change to parse as a float after regex.

## Issues addressed

#189 

## Screenshots

N/A but can go to any course/review page to sort


j-alicia-long (approved):
Checked it out, looks good!

PR #217, opened 2020-11-14
Revert PR#211 vote fix

Bartleby2718:
# Status
Done

# What I did
Basically reverted PR #211 and fixed the bug that existed at that point with minimal changes

## Issues addressed
Reverts #211

## Screenshots
image

Notice that there is now a new `id` attribute and no `href`.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,0 @@
-tcf_website/static/reviews/review.js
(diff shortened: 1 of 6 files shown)


j-alicia-long (approved):
Checked it out and it works! Thanks for catching the bug.

PR #218, opened 2020-11-14
Added latest semester to old professor classes

jackliu612:
# Status
Done, In progress, etc

# What I did

Fixed a bug where the last semester taught by old professors aren't showing in the course list

Modified models to add a last_taught method to instructors as well as changes to browse view and course template to display and sort. 

## Screenshots

Old:
image

New:
image

### DIFF
--- a/tcf_website/models/models.py
+++ b/tcf_website/models/models.py
@@ -199,6 +199,15 @@ class Instructor(models.Model):
         # implies (collecting Sections instead of Courses); work in progress
         return Section.objects.filter(instructors=self)
 
+    def last_taught(self, course):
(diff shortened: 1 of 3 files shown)


Bartleby2718 (commented):
The code looks good, but can u test the overhead by comparing load times for several different pages before and after this commit? Especially classes that have been taught by many different professors. One I can think of is MATH 3351.

jackliu612:
@Bartleby2718 I checkout some performance between classes and unfortunately it can be very bad. 
Here's silk on the dev site:
image

and here it is on this branch:
image

The classes are ECON 2010 and MATH 3351 and it looks like there are about double the amount of queries

PR #237, opened 2020-11-17
Lous list script

j-alicia-long:
# Status
Done

# What I did
Generated table for Lou's List links to work correctly, and wrote documentation on how to do it!

## Issues addressed
[UVA Lou's List](louslist.org) links individual courses to tCF course pages, and it requires a table mapping course IDs to the IDs in our URLs (mnemonic, number, database id). They are then fetched at https://thecourseforum.com/course/[course_id].


alex-shen1:
Do we need to keep a copy of the CSV tracked in the repo if it just corresponds to data that's in the DB? Seems like it'd be better to just generate this once per semester as needed and send it off to Lou then. We don't actually need the file any other time, so it just adds bloat that probably isn't necessary.

Also, nice job figuring this out/writing up good docs! I think at some point we'll want to put this in a wiki like @Bartleby2718 was suggesting instead of having extra instruction Markdown files (and random code comments in other scripts :/) in the repo, but it'll work for now.

j-alicia-long:
> Do we need to keep a copy of the CSV tracked in the repo if it just corresponds to data that's in the DB? Seems like it'd be better to just generate this once per semester as needed and send it off to Lou then. We don't actually need the file any other time, so it just adds bloat that probably isn't necessary.
> 
> Also, nice job figuring this out/writing up good docs! I think at some point we'll want to put this in a wiki like @Bartleby2718 was suggesting instead of having extra instruction Markdown files (and random code comments in other scripts :/) in the repo, but it'll work for now.

I just pushed a new commit reorganizing scripts & DB stuff into a subfolder. @alex-shen1 feel free to go in and delete any unneeded or outdated data files!

Bartleby2718:
Made the command more concise, updated the doc accordingly, and moved/renamed/deleted some files.
@alex-shen1 @jackliu612 I would like your reviews since I didn't work on migration from tCF 1 to tCF 2.

db/LousListScript.md
```diff
@@ -0,0 +1,18 @@
+# Lou's List Script
+[UVA Lou's List](louslist.org) links individual courses to tCF course pages, and it requires a table mapping course IDs to the IDs in our URLs (mnemonic, number, database id). They are then fetched at https://thecourseforum.com/course/[course_id].
+
+Run the following commands to generate a table for Lou's List using our database:
+
+0. Make sure you have your database set up correctly. It is likely that you ran `docker-compose down` to start with a clean slate.
```
alex-shen1, line 6: Minor wording clarification —  this means you're supposed to run `docker-compose down` and then `docker-compose up` to rebuild everything from scratch, right?

db/LousListScript.md
```diff
@@ -0,0 +1,18 @@
...
+
+1. Open the psql terminal in the tCF database Docker container with this command*:
```
alex-shen1, line 8: Technically after making it a oneliner we're not "opening the psql terminal" anymore, as far as I know

db/LousListScript.md
```diff
@@ -0,0 +1,18 @@
...
+    ```
+    docker exec -it tcf_db psql -U tcf_django tcf_db -f app/export_mapping_table.sql
```
alex-shen1, line 10: Connected to above; I've tested and it works without the `-it` flag (as it's just running one thing)

alex-shen1: Removing `-it` also makes it work on Git Bash on Windows!

alex-shen1 (commented):
Nice work! Have a few minor nitpicks with the instructions, but the procedure does work as expected. Also two separate comments that aren't directly related to code:
- Currently this branch has the Lou's export tracked — I've said before that since the export is a once-per-semester thing that I think we should gitignore it, but do you have any opinions on this?
- I believe the `mysql2sqlite` and `mysql2sqlite2` were scripts that we only used to migrate tCF1 data over, so I _think_ it should be safe to delete them unless we plan to go back and do something else with the data? I can't exactly imagine a scenario where it would be helpful, but perhaps it'd be good to keep it around so we can remember what we did if we ever need to. @jackliu612 worked more on that aspect than I did, so maybe he can chime in with some thoughts.

db/LousListScript.md
Bartleby2718: Good catch! Fixed in https://github.com/thecourseforum/theCourseForum2/pull/237/commits/6d71c9265e5ae72ff4aebc7cbad263c729e2b81d

db/LousListScript.md
Bartleby2718: Fixed in https://github.com/thecourseforum/theCourseForum2/pull/237/commits/6d71c9265e5ae72ff4aebc7cbad263c729e2b81d. Thanks for testing it out as I couldn't test it because I don't have it installed.

db/LousListScript.md
Bartleby2718: Clarified in https://github.com/thecourseforum/theCourseForum2/pull/237/commits/6d71c9265e5ae72ff4aebc7cbad263c729e2b81d

db/LousListScript.md
Bartleby2718: @alex-shen1 What if you remove just the `-t` flag? I was running another command, and it seems like `-i` is necessary in that case.

alex-shen1: Yeah, it works if you do `-i` or `-t`, but breaks only if you do `-it`. Weird; I think it's something with the terminal emulator Git Bash uses on Windows (weirdly enough, if you run Git Bash from Pycharm `-it` works fine)

alex-shen1 (approved):

PR #239, opened 2020-11-18
Added Google Analytics tags

julesfll:
# Status
In progress/Done (depending on how many tags we want)

# What I did
Added a Google Analytics tag for upvotes and downvotes. This will let us see when people are upvoting and downvoting reviews on Google Analytics. It's really a trial for more interesting events such as writing reviews (so we can figure out how to encourage people to write reviews)

Having metrics like these will allow us to measure the impact of features we add.
## Screenshots
image


vn6:
Looks like there's some eslint problems with the travis build. I'm also wondering if there's a simpler way to do this so we don't have to add tags to a bunch of code (i.e. in google analytics, it seems to be tracking getting search results without the tag), seems like using google tag manager might be an option from a quick search online.

julesfll:
> Looks like there's some eslint problems with the travis build. I'm also wondering if there's a simpler way to do this so we don't have to add tags to a bunch of code (i.e. in google analytics, it seems to be tracking getting search results without the tag), seems like using google tag manager might be an option from a quick search online.
> 
> Good idea though, seems like this would be really useful!

It's worth looking into Tag Manager for sure. Although that may be an additional learning curve whereas since we know how to code it could be easier to put them in manually. I do see the value in decoupling the code though

PR #240, opened 2020-11-19
Refactor and improve the speed of /course/{course_id}

### LINKED ISSUE #243: LAST TAUGHT not shown for "old" instructors
image

Bartleby2718:
# Status
Done

# What I did
- added `semester_last_taught` to "old" professors
- refactored a Django view (`course_view`) to reduce unnecessary database access
  - changed `course.html` accordingly

## Issues addressed
This fixes #243 and supercedes #218.

## Screenshots
image
As you can see, the performance difference between and after the refactoring is dramatic. Note that the requests in the red box do not even include `semester_last_taught` for "old" professors, while the requests in the green box do.

### DIFF
--- a/tcf_website/templates/course/course.html
+++ b/tcf_website/templates/course/course.html
@@ -37,16 +37,10 @@
             <br/>
 
             <ul class="instructor-list">
-                {% for i in recent_instructors %}
-                    <li class="instructor">
+                {% for i in instructors %}
+                    <li class="instructor {% if i.semester_last_taught != latest_semester %}old{% endif %}">
                         {% url 'course_instructor' course_id=course.pk instructor_id=i.pk as link %}
-                        {% include "common/rating_card.html" with title=i.full_name link=link rating=i.rating difficulty=i.difficulty gpa=i.gpa last_taught=course.semester_last_taught %}
-                    </li>
-                {% endfor %}
-                {% for i in old_instructors %}
-                    <li class="instructor old">
-                        {% url 'course_instructor' course_id=course.pk instructor_id=i.pk as link %}
-                        {% include "common/rating_card.html" with title=i.full_name link=link rating=i.rating difficulty=i.difficulty gpa=i.gpa %}
+                        {% include "common/rating_card.html" with title=i.full_name link=link rating=i.rating difficulty=i.difficulty gpa=i.gpa last_taught=i.semester_last_taught %}
                     </li>
                 {% endfor %}
             </ul>
--- a/tcf_website/views/browse.py
+++ b/tcf_website/views/browse.py
@@ -3,7 +3,7 @@
 """Views for Browse, department, and course/course instructor pages."""
 import json
 
-from django.db.models import Avg
+from django.db.models import Avg, Max, Q
 from django.shortcuts import render
 from django.urls import reverse
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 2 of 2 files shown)


tcf_website/templates/course/course.html
```diff
@@ -37,16 +37,10 @@ <h2>{{ course.title }}</h2>
             <br/>
 
             <ul class="instructor-list">
-                {% for i in recent_instructors %}
-                    <li class="instructor">
+                {% for i in instructors %}
+                    <li class="instructor {% if i.semester_last_taught != latest_semester %}old{% endif %}">
```
alex-shen1, line 41: Nice refactor!

tcf_website/views/browse.py
```diff
@@ -75,36 +75,26 @@ def course_view(request, course_id):
...
+        .filter(section__course=course).distinct()\
+        .annotate(
+            gpa=Avg('courseinstructorgrade__average',
+                    filter=Q(courseinstructorgrade__course=course)),
+            difficulty=Avg('review__difficulty', filter=Q(review__course=course)),
+            rating=(
+                Avg('review__instructor_rating', filter=Q(review__course=course)) +
+                Avg('review__enjoyability', filter=Q(review__course=course)) +
+                Avg('review__recommendability', filter=Q(review__course=course))
+            ) / 3,
```
alex-shen1, line 88: We have an `average_rating_for_course(self,course)` method in the Instructor model that basically does this — it might be better for readability if we just used that instead. Is there a performance improvement by doing this instead?

tcf_website/views/browse.py
```diff
@@ -75,36 +75,26 @@ def course_view(request, course_id):
...
+            semester_last_taught=Max('section__semester',
+                                     filter=Q(section__course=course)),
+        )
+    # Note: Wanted to use .annotate() but couldn't figure out a way
+    # So created a dictionary on the fly to minimize database access
+    semesters = {s.id: s for s in Semester.objects.all()}
+    for instructor in instructors:
+        instructor.semester_last_taught = semesters.get(
+            instructor.semester_last_taught)
```
alex-shen1, line 97: Instead of iterating through the loop to check against `latest_semester` (which, to my understanding, is currently an instance of Semester), wouldn't it be easier to make `latest_semester` also just an ID to check against?

tcf_website/templates/course/course.html
Bartleby2718: Actually, the idea of wrapping `old` in an `if` tag was taken from the department page. I'm just a copycat. 😄

tcf_website/views/browse.py
Bartleby2718: `latest_semester` is used in the template https://github.com/thecourseforum/theCourseForum2/blob/a913e825459c23b09df9daa5d452a3aa6d6910dc/tcf_website/templates/course/course.html#L18 so we need to access the `Semester` table anyway.

alex-shen1: It might be better to compare `i.semester_last_taught` against `lastest_semester.id` (not sure if that's the right way to do it) in the template then?

alex-shen1: update: can confirm you can access the id with `latest_semester.id` — definitely think that's the way to go instead of changing fields

tcf_website/views/browse.py
Bartleby2718: The method you're referring to works for an *instance* of `Instructor`, but not for a *queryset* of `Instructor`s. 
image
We need to work with querysets so that we don't access the database for every instance.

tcf_website/views/browse.py
Bartleby2718: Okay, let's start from the basics to make sure we're on the same page.
1. Before this part of the code, I have as `semester_last_taught=Max('section__semester', filter=Q(section__course=course))`, which is an `int`. (I couldn't figure out a way to make it an instance of `Semester`, and I don't think it's possible.)
2. However, what the template needs is an instance of `Semester` (or a string representation thereof) for each `Instructor`. That's why I need to access the database again. However, accessing the database within a loop would be very inefficient, so instead I created a `Dict[int, Semester]` in line 94.
3. Going back to your first comment, I'm not sure why you brought up `latest_semester`, which is the most recent semester at the moment. What we need is the most recent semester *for each `Instructor` teaching a given course*, which I denote as `semester_last_taught`.
4. As I said in 1, I only have the ID (which is an `int`) of the semester last taught. Line 96 is converting the `int` to an instance of `Semester` for the reason mentioned in 2, which is possible because Python is dynamically typed.
5. Just so we're on the same page, the conversion in 4 does change a field of an `Instructor` instance, but we didn't call `.save()`, so the change does not propagate to the database (which should fail anyway).

With that said, can you clarify why `latest_semester` was mentioned?

alex-shen1: Thanks for the clarification — I overlooked one of the sections of the code where you pass the `Semester` instance itself into the `rating_card.html` template... I saw the `{% if i.semester_last_taught != latest_semester %}` refactor and assumed that this could be simplified with just using IDs. Sorry for the confusion!

alex-shen1 (approved):

PR #242, opened 2020-11-19
Improve the performance of reviews without client-side rendering

### LINKED ISSUE #241: Reviews page is very slow
Reported by @jackliu612 
image

Bartleby2718:
# Status
Done

# What I did
Initially, I was going to use client-side rendering to solve this issue, but while I was working on it, I discovered that a significant amount of performance gain can be achieved just by fixing some ORM code. That's what I did in this PR, but let me know if this is still slow. I think 3 seconds in the worst case is pretty reasonable.

## Issues addressed
Closes #241

## Screenshots
image


tcf_website/models/models.py
```diff
@@ -690,23 +700,24 @@ def downvote(self, user):
...
-                except Vote.DoesNotExist:
-                    review.user_vote = 0
-
-        return reviews
+            reviews = reviews.annotate(
+                user_vote=models.functions.Coalesce(
+                    models.Sum('vote__value',
+                               filter=models.Q(vote__user=user)),
+                    models.Value(0)
+                ),
```
alex-shen1, line 718: Just for my own curiosity: _why_ does this work? I assume it's just because it does it in one SQL query instead of a whole bunch?

alex-shen1 (approved):
Works fine on my local build!

tcf_website/models/models.py
Bartleby2718: The original loop accesses the database for each review, whereas this query accesses the database only once. Also, this does not cause any exceptions--they incur a lot of overhead.

PR #244, opened 2020-11-19
Add copyright year correctly & dynamically

### LINKED ISSUE #219: Copyright year
Our sidebar takes the copyright year from most recent semester, so during the fall when we import spring data, the year will be prematurely set to the next year.

Bartleby2718:
# Status
Done

# What I did

## Issues addressed
Closes #219

## Screenshots
image

### DIFF
--- a/tcf_website/templates/base/sidebar.html
+++ b/tcf_website/templates/base/sidebar.html
@@ -69,7 +69,7 @@
   <div class="footer text-center small">
     <a href="{% url 'privacy' %}">Privacy</a> |
     <a href="{% url 'terms' %}">Terms</a>
-    <p>theCourseForum <br> &copy; {{ LATEST_SEMESTER.year }} v1.0.0</p>
+    <p>theCourseForum <br> &copy; <span class="year"></span> v1.0.0</p>
     {% if DEBUG %}
     <div class="bg-tcf-blue pb-3">
       <hr class="border-white">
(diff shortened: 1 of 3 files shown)


tcf_website/templates/base/sidebar.html
```diff
@@ -69,7 +69,7 @@ <h4><i class="fa fa-user fa-fw"></i></h4>
   <div class="footer text-center small">
     <a href="{% url 'privacy' %}">Privacy</a> |
     <a href="{% url 'terms' %}">Terms</a>
-    <p>theCourseForum <br> &copy; {{ LATEST_SEMESTER.year }} v1.0.0</p>
+    <p>theCourseForum <br> &copy; <span class="year"></span> v1.0.0</p>
```
alex-shen1, line 72: This is outside the scope of this PR so I won't have this as something that blocks merging, but what "version" are we really on anyways? tCF2 1.0? Right now we just have "theCourseForum" on the page, but the old version was "theCourseForum 3.0.1", so it's a bit weird that we're effectively going backwards. Should probably discuss this at meeting, I'll add it to the agenda.

alex-shen1 (approved):
Nice work!

julesfll (approved):
Looks good!

tcf_website/templates/base/sidebar.html
Bartleby2718: We should probably follow https://semver.org/, but I was wondering the same thing...

PR #246, opened 2020-11-19
URL after login

### LINKED ISSUE #245: URL gets weird after login
Users are redirected to the browse page after login, but the URL is `/account/profile`
image

Bartleby2718:
# Status
Done, In progress, etc

# What I did
- Changed `SOCIAL_AUTH_LOGIN_REDIRECT_URL` to the browse page
- Removed an unnecessary path from `tcf_website/urls.py`

## Issues addressed
Closes #245

## Screenshots
image

### DIFF
--- a/tcf_core/settings.py
+++ b/tcf_core/settings.py
@@ -12,6 +12,7 @@ https://docs.djangoproject.com/en/3.0/ref/settings/
 """
 
 import os
+from django.urls import reverse_lazy
 from django.contrib.messages import constants as messages
 import environ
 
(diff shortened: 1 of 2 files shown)


tcf_core/settings.py
```diff
@@ -152,6 +153,7 @@
 SOCIAL_AUTH_GOOGLE_OAUTH2_KEY = env.str('SOCIAL_AUTH_GOOGLE_OAUTH2_KEY')
 SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET = env.str('SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET')
 SOCIAL_AUTH_GOOGLE_OAUTH2_WHITELISTED_DOMAINS = ['virginia.edu']
+SOCIAL_AUTH_LOGIN_REDIRECT_URL = reverse_lazy('browse')
```
julesfll, line 156: What does `reverse_lazy` do as opposed to `reverse`? Just curious

julesfll (approved):
Looks good! Great work

tcf_core/settings.py
Bartleby2718: I actually tried `reverse` after initially using `reverse_lazy` (to save 10 characters), but it didn't work. 
https://docs.djangoproject.com/en/3.1/ref/urlresolvers/#reverse-lazy
> A lazily evaluated version of reverse().
> 
> It is useful for when you need to use a URL reversal before your project’s URLConf is loaded. Some common cases where this function is necessary are:
> 
> - providing a reversed URL as the url attribute of a generic class-based view.
> - providing a reversed URL to a decorator (such as the login_url argument for the django.contrib.auth.decorators.permission_required() decorator).
> - providing a reversed URL as a default value for a parameter in a function’s signature.

PR #253, opened 2020-11-22
Fix #252 Duplicate data in the Instructor API

### LINKED ISSUE #252: Duplicate data in the Instructor API
Instructors have duplicates, need to be distinct

Kinda blocking #67

Bartleby2718:
# Status
Done

# What I did
Added a filter that returns only a unique set of `Instructor`s

## Issues addressed
Fixes #252

## Screenshots
Before the fix (count=25):
image
After the fix (count=9):
image


j-alicia-long (approved):
Looks good, thanks for fixing this really quickly! Merge it in once Travis passes.

PR #257, opened 2020-11-22
Add recency param to Instructor API

### LINKED ISSUE #254: Add recency param to Instructor API
allows you to fetch results from the last X years

Note: course param would be required for this^

Bartleby2718:
# Status
Done

# What I did
- Added `recent5years` param to Instructor API
  - Not sure about the name `recent5years` (maybe just `recent`?)
  - Not sure about the behaviors described below as they are inconsistent with those of `CourseViewSet` (which returns a filter queryset regardless of the value of `recent5years`)--can argue for both (simplicity vs. clarity)
  - Maybe merge this PR first and have another PR to address the two issues above?

## Issues addressed
Closes #254

## Screenshots
1. `section__course` only
image
2. valid `section_course` and `recent5years=false`
image
3. valid `section_course` and invalid `recent5years`
image
4. valid `section_course` and valid `recent5years` (i.e. `true`)
image


tcf_website/api/filters.py
```diff
@@ -1,12 +1,20 @@
...
-from django_filters import FilterSet, NumberFilter
-from ..models import Instructor
+from django_filters import FilterSet, BooleanFilter, NumberFilter
+from ..models import Instructor, Semester
 
 
 class InstructorFilter(FilterSet):
     """Filter for Instructor"""
     section__course = NumberFilter(field_name='section__course', distinct=True)
+    recent5years = BooleanFilter(method='filter_recent5years', distinct=True)
```
alex-shen1, line 9: Definitely think calling the param `recent` makes more sense from a legibility perspective — just add a comment somewhere in the code / documentation so that it's clear that we know what "recent" means.

alex-shen1 (commented):
Just needs small field name change and fixing pylint errors and it should be good to go!

tcf_website/api/filters.py
j-alicia-long: I'm gonna make this change (and fix pylint) and merge it in since it's blocking review form rn

j-alicia-long:
Checked it out, works fine! For some reason I'm getting this error on the DRF dashboard page (http://localhost:8000/api/instructors/)

image

GET request still works fine though, so it shouldn't impact the functionality.

j-alicia-long (approved):

PR #282, opened 2020-11-26
Fix semester casing in write-review page

Bartleby2718:
# Status
Done

# What I did
- Change how `season` is cased in the API. 
  - Note that other APIs are also affected because `SemesterSerializer` is often nested in another serializer.

## Screenshots
Before:
image

After:
image

### DIFF
--- a/tcf_website/api/serializers.py
+++ b/tcf_website/api/serializers.py
@@ -6,6 +6,11 @@ from ..models import (Course, Department, School, Instructor, Semester,
 
 class SemesterSerializer(serializers.ModelSerializer):
     """DRF Serializer for Semester"""
+    season = serializers.SerializerMethodField()
+
+    def get_season(self, obj):
+        return obj.season.title()
+
     class Meta:
         model = Semester
         fields = '__all__'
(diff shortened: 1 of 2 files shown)


alex-shen1 (dismissed):
Works for me!

tcf_website/api/serializers.py
```diff
@@ -6,6 +6,13 @@
 
 class SemesterSerializer(serializers.ModelSerializer):
     """DRF Serializer for Semester"""
+    season = serializers.SerializerMethodField()
+
+    def get_season(self, obj):
+        """Change the `season` field to TitleCase (or PascalCase)"""
+        # pylint: disable=no-self-use
```
alex-shen1, line 13: One small thing — I think most of the other pylint disables in the codebase have been on the first (few) lines of each file as opposed to where the would-be offending code is.

Bartleby2718: Fixed in 7c18b253bc988342486c684df611cc6978c9debc!

alex-shen1 (approved):

PR #284, opened 2020-11-26
Upgrade Python version

### LINKED ISSUE #283: Upgrade Python version
- Python 3.9 was out last month, and [it has been a little over a year since Python 3.8 was released](https://www.python.org/downloads/release/python-380/).
- Both [GAE](https://cloud.google.com/appengine/docs/standard/python3/runtime#python_3_versions) and [Heroku](https://devcenter.heroku.com/articles/python-support#supported-runtimes) support Python 3.8.
- I think now it is safe to use a newer version of Python, which will allow us to new features like [assignment expression](https://www.python.org/dev/peps/pep-0572/).

Bartleby2718:
# Status
Done

# What I did
- Bumped Python version to 3.8
- See #283 for the rationale behind this PR.

## Issues addressed
- Closes #283


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.6.5
+FROM python:3.8
```
alex-shen1, line 1: [Docker Hub](https://hub.docker.com/_/python) appears to have a `3.8.6` version as opposed to simply `3.8` — considering you explicitly refer to `python-3.8.6` in `runtime.txt`, would it be possible/better to use this instead?

Bartleby2718: The thing is, we cannot specify the `patch` version (the number after the second period) for Google App Engine, whereas we do need to specify it for Heroku. We can do either for Docker, so it will be inconsistent in any case, and I can't really argue for either side. To play devil's advocate, if there was a bug fixed in a more recent patch of Python, wouldn't it be easier to figure it out if the bug could be found either in the local environment or on Heroku? Although I think such a case would be very unlikely, I wouldn't suspect the Python patch version could be an issue if the bug could only be found in the production environment.

Bartleby2718: @alex-shen1 But then you will stay in the team for a couple more years, so I specified the patch version too. I think my scenario was pretty unrealistic anyway.

alex-shen1 (approved):

j-alicia-long (commented):
Looks good to me! When do you recommend we upgrade to Python 3.9?

Bartleby2718:
@j-alicia-long Maybe when GAE starts supporting Python 3.9? Right now it doesn't: 
> The Python 3 runtime supports Python 3.7 and Python 3.8, and it uses the latest stable release of the version you choose. App Engine automatically updates to new release versions upon app deployment, but it will not automatically update the minor version.
> 
> For example, your application might be deployed at Python 3.7.0 and later automatically updated to Python 3.7.1, but it will not be automatically updated to Python 3.8.0.
https://cloud.google.com/appengine/docs/standard/python3/runtime#python_3_versions.

PR #291, opened 2020-12-01
Statistics on My Review page

### LINKED ISSUE #267: Statistics on My Review page

Bartleby2718:
# Status
Done

# What I did
- Added previous statistics (number of reviews, number of upvotes, and average review rating) to my reviews page for parity

## Issues addressed
- Closes #267

## Screenshots
- tCF 1.0
image
- Mine
image
- current tCF 2.0
image


j-alicia-long (commented):
Nice work! 

Wondering if the total upvotes includes only your upvotes, or also downvotes (in terms of contributions?)
Or is it like...the total number of times your own reviews have been upvoted?
Might want to add a tooltip to comment that :)

tcf_website/static/reviews/review_stats.css
```diff
@@ -0,0 +1,18 @@
...
+}
+
+.large-stats {
+    padding-top: 15px;
+    padding-bottom: 10px;
+    background-color: white;
+    text-align: center;
+    display: block;
+    margin-left: 1px;
+    width: 33.33%
```
j-alicia-long, line 13: Do you need the `width: 33.33%` if you're already using bootstrap columns?

tcf_website/views/profile.py
```diff
@@ -51,4 +53,21 @@ def profile(request):
...
+    # Get other statistics
+    other_stats = User.objects.filter(id=request.user.id).aggregate(
+        total_reviews_written=Count('review'),
+        average_review_rating=(
+            Avg('review__instructor_rating') +
+            Avg('review__enjoyability') +
+            Avg('review__recommendability')
+        ) / 3,
+    )
+    # Merge the two dictionaries (NOTE: can be simplified in Python 3.9)
```
j-alicia-long, line 69: hmm any recommendations on when we should upgrade python?

tcf_website/templates/reviews/review_stats.html
```diff
@@ -0,0 +1,16 @@
+<div class="large-stats-container container-lg">
+    <div class="d-flex">
+        <div class="col-xs-4 large-stats">
+            <h3>Total Reviews Written</h3>
```
j-alicia-long, line 4: For verbosity reasons might want to remove the redundant "Review" word from each stat panel, since you already have the "My Reviews" header right above that

tcf_website/views/profile.py
Bartleby2718: Moving the discussion to https://github.com/thecourseforum/theCourseForum2/pull/284#issuecomment-737737870

tcf_website/templates/reviews/review_stats.html
Bartleby2718: Yeah, I also thought it was kinda redundant but used the original wording for parity.

Bartleby2718:
@j-alicia-long Thanks for the review!

I think using a better phrase would be better than a tooltip. Is `Total Upvotes Received` still ambiguous?

tcf_website/static/reviews/review_stats.css
Bartleby2718: I'm not really good at frontend, but the `width` was there in the previous code and it looks like this without the line:
image
Seems like it was there for a reason in the first place?

j-alicia-long: ehhh that's strange 
lol just leave it then

j-alicia-long:
> @j-alicia-long Thanks for the review!
> 
> I think using a clearer phrase would be better than a tooltip. Is `Total Upvotes Received` still ambiguous?

hmm I would keep it as a two-word header to match the others (just a nitpick thing)

I'm gonna go in and try to mess around with the css to attempt to simplify some stuff

j-alicia-long:
Cleaned up CSS and made it a lil more responsive!

image
image

tcf_website/views/profile.py
```diff
@@ -3,10 +3,12 @@
 from django.shortcuts import render
 from django.contrib.auth.decorators import login_required
 from django.contrib import messages
+from django.db.models import Avg, Count, Q
 from django import forms
 from django.forms import ModelForm
 from django.http import HttpResponseRedirect
-from ..models import User
+from .browse import safe_round
```
alex-shen1, line 10: Little bit unsure if it's a good idea to import helper functions from another view, but I can't think of a better way to do it — only alternatives would either be duplicating the function or creating some new file they can both import from, which don't strike me as particularly good options.

alex-shen1 (approved):
Looks great!

tcf_website/views/profile.py
Bartleby2718: Do you think it's better to create `utils.py`, which keeps all the helper functions?

j-alicia-long: > Do you think it's better to create `utils.py`, which keeps all the helper functions?

That sounds like a good idea to me

Bartleby2718: Let's discuss this in https://github.com/thecourseforum/theCourseForum2/issues/303 since it also has to do with `emdashOrTwoDecimals` in `tcf_website/static/department/loadMoreCourses.js`. Marking the conversation as resolved.

vn6 (approved):
Looks good!

tcf_website/views/profile.py
```diff
@@ -51,4 +53,21 @@ def profile(request):
...
+        total_review_upvotes=Count('vote', filter=Q(vote__value=1)),
+    )
+    # Get other statistics
+    other_stats = User.objects.filter(id=request.user.id).aggregate(
+        total_reviews_written=Count('review'),
+        average_review_rating=(
+            Avg('review__instructor_rating') +
+            Avg('review__enjoyability') +
+            Avg('review__recommendability')
+        ) / 3,
```
j-alicia-long, line 67: Not super important but I wonder if this logic would be better as a function in the User model

Bartleby2718: There was [a similar discussion](https://docs.djangoproject.com/en/3.1/topics/db/managers/) before, but the reason why our code was very slow in general is that we used functions (for individual instance) and iterated over querysets to invoke those functions. A faster and more DB-friendly approach is to use `JOIN`s and aggregation, which translates to annotations in Django ORM. I've been looking into [managers and querysets](https://docs.djangoproject.com/en/3.1/topics/db/managers/) to pull out repeated annotations, but it has low priority because it's just for refactoring.

Bartleby2718: If this is the only remaining issue, it seems like we can squash-merge this PR?

j-alicia-long: Gotcha, makes sense! Will merge for you

PR #292, opened 2020-12-01
Lous suggestion

Bartleby2718:
# Status
Done

# What I did
- Rename the subfolder
- Renamed/removed files
- Mount volumes to the DB container
- Create a separate SQL file for the Lou's List script

## Screenshots
image


FOR-DEVS.md
```diff
@@ -58,14 +58,16 @@ If you are part of theCourseForum engineering team, follow the instructions belo
...
 4. In the `theCourseForum2/` directory, run `docker-compose up` to start your Docker containers (tCF Django app and database)
-5. Download a copy of the database from [Google Drive](https://drive.google.com/open?id=1ubiiOj-jfzoBKaMK6pFEkFXdSqMuD-22)
+5. Download a copy of the database from [Google Drive](https://drive.google.com/file/d/1S9GNFydvw6ZZW0AGHy6zyTLl-pvyIT8Q/view?usp=sharing)
     - put this into the base `theCourseForum2/` directory
-6. While your container is still running, open a second terminal, cd into `theCourseForum2/`, and run this script to set up your database*:
-    - `./load_data ${DB}` 
-    - where `${DB}` should be replaced with the latest DB file (e.g. `./load_data april7.sql`)
+6. Once the Django server is done loading (i.e. when you see `tcf_django | Watching for file changes with StatReloader`), open a second terminal, cd into `theCourseForum2/`, and run the following two lines of script to set up your database, one at a time:
+    ```
+    docker exec -i tcf_db psql tcf_db -U tcf_django -c "DROP SCHEMA public CASCADE; CREATE SCHEMA public;"
```
j-alicia-long, line 65: did you rename the database to `public`?

j-alicia-long (approved):
Looks good! Great job consolidating the `load_data` script into a one-liner. 

The `db` directory name also seems more intuitive since we don't seem to have any non-db related scripts yet.

FOR-DEVS.md
Bartleby2718: @j-alicia-long  No, I checked and all our tables belong to a single schema named `public`, which I think is the default value. I believe there are multiple schemas in a single database and multiple tables in a single schema. I found this method here: https://stackoverflow.com/questions/3327312/how-can-i-drop-all-the-tables-in-a-postgresql-database.
This is what you see when you run `docker exec -ti tcf_db psql tcf_db -U tcf_django -c "select * from pg_tables;"`:
image
Seemed like a good trick for us.

j-alicia-long: @Bartleby2718 Hmm that's interesting since our DB setup/reset docs says to drop & recreate the `tcf_db` database?
image

Bartleby2718: @j-alicia-long Just so we're on the same page, the hierarchy looks like this (`1---*` means one-to-many relationship):
> PostgreSQL server running on the `tcf_db` container
> - 1---* database
>   - 1---* schema
>      - 1---* table

Should I have written `schema` in the doc? 🤔

Issue #293, opened 2020-12-02
Multiple CourseInstructorGrade instances for a single (Course, Instructor) pair


Bartleby2718:
- Bug Report
  > - URL: https://thecourseforum.com/course/7421/3761
  > - Description: 
  > Hi! Are the grade distributions going to be fixed before course registration? Many classes that previously had the grade pie 
chart > filled in have not had it since the update. Thanks!
  > - Categories: [Data Error]
- Explanation
  - This is because a part of the current code assumes that `CourseInstructorGrade`  is unique for each (`Course`, `Instructor`) pair. It _should_ be, but it _currently is not_, because the underlying data is not so clean. Need to overhaul `loade_grades.py`.

alex-shen1:
Also results in this error on prod for duplicate instances (i.e. John Hott for CS 2110):
![image](https://user-images.githubusercontent.com/55100084/101476614-0aeb6a80-391c-11eb-9953-162a4f51637c.png)

alex-shen1:
@Bartleby2718 correct me if I'm wrong but I believe this issue is fixed by the merged PR; for the future, don't forget to add "Closes #{issue}" to initial PR descriptions so we can automatically close issues when they get merged

Bartleby2718:
@alex-shen1 If I remember correctly, this issue was resolved temporarily, but there is no guarantee that this won't happen again. That's why I was hesitant to close this even though I knew I could close this using the keyword, but I guess we can keep it closed for now.

PR #306, opened 2020-12-05
Fix grades

Bartleby2718:
# Status
Done

# What I did
- No big changes in the code except discarding middle names, but creating a PR for the SQL scripts to be merged ASAP.
- Will continue to improve `load_grades.py` in the coming weeks

## Issues addressed
- This is a quick and dirty fix to #293 but not a permanent one. However, running `db/link_courseinstructorgrade_to_instructor.sql` will allow us to restore around 80% of the missing `CourseInstructor` instances by giving it a proper `Instructor` foreign key.

## Screenshots
image
4034 -> 834 is roughly an 80% decrease in the number of cases where `CourseInstructor` was not linked to `Instructor`.

## How to Test 
- Start by running `docker exec -ti tcf_db psql tcf_db -U tcf_django -c "SELECT first_name, last_name, course_id, COUNT(*) AS num FROM tcf_website_courseinstructorgrade WHERE course_id IS NOT NULL GROUP BY first_name, last_name, course_id ORDER BY num DESC;"` (and then exit by entering `\q` + return). You should see a few rows whose `num` columns are not 1 on the top.
- Run `docker-compose down && docker-compose up`, wait until the Django server is up and running, and then run the following commands in a separate terminal, one at a time in the following order:
```
docker exec -i tcf_db psql tcf_db -U tcf_django -c "DROP SCHEMA public CASCADE; CREATE SCHEMA public;"
docker exec -i tcf_db psql tcf_db -U tcf_django < launch.sql
// In the production environment, we will start from here, creating a backup beforehand to be safe.
// Of course, we will only run the part after `docker exec -i {containerName}` since we don't use Docker on GAE or DO.
// The SQL scripts are not in the DO, so you can copy/paste the content instead of using the `-f` flag.
docker exec -i tcf_django python3 manage.py load_grades ALL_DANGEROUS 
docker exec -i tcf_django python3 manage.py fix_multiple_objects_returned
docker exec -i tcf_db psql tcf_db -U tcf_django -f app/link_courseinstructorgrade_to_instructor.sql
docker exec -i tcf_db psql tcf_db -U tcf_django -f app/add_missing_instructor_emails.sql
```
- Now everything _should_ be fixed. Again, run `docker exec -ti tcf_db psql tcf_db -U tcf_django -c "SELECT course_id, instructor_id, COUNT(*) AS num FROM tcf_website_courseinstructorgrade cig WHERE course_id IS NOT NULL AND instructor_id IS NOT NULL GROUP BY course_id, instructor_id ORDER BY num DESC;"` (and then entering `q` + return). You should only see 1 in the `num` column now.


tcf_website/management/commands/load_grades.py
```diff
@@ -235,11 +244,11 @@ def load_dict_into_models(self):
                 'subdepartment': row[0],
                 'number': row[1],
                 'first_name': row[2],
-                'middle_name': row[3],
-                'last_name': row[4],
-                'email': row[5],
+                'middle_name': '',  # for backward compatibility
```
jackliu612, line 247: Was middle name removed from the loading because recent files don't include it anymore?

db/add_missing_instructor_emails.sql
```diff
@@ -0,0 +1,9 @@
+-- Copy `email`s from CourseInstructorGrade to Instructor
+-- assuming instructor names are unique
```
jackliu612, line 2: We make this assumption in our models and migration so I think it's ok but this might be the source of quirks in our data. Also, if there are duplicates would it just replace with the first email it finds under the name?

jackliu612 (commented):
Overall it looks good and the 80% reduction is great. I didn't have that much time to look through all of it but I left just two (mostly notes) comments down.

tcf_website/management/commands/load_grades.py
Bartleby2718: No, because 1) there are inconsistencies in the CSV files and 2) `Instructor` table doesn't have middle name, so I figured we could drop it entirely.

jackliu612: I think that's fair

jackliu612:
Also, do these scripts/sql commands need to be run each time load_grades.py is run as well?

Bartleby2718:
@jackliu612 Actually running `load_grades` for a single semester deletes all existing data, so I need to fix that first. After that, I'll think about whether this script should be incorporated to `load_grades` or run separately.

tcf_website/management/commands/fix_multiple_objects_returned.py
```diff
@@ -0,0 +1,113 @@
...
+    """This fixes existing `MultipleObjectsReturned` errors"""
+
+    help = 'Fix `MultipleObjectsReturned` errors that were caused by inconsistent email data'
+
+    def handle(self, *args, **options):
+        self.verbosity = options['verbosity']
+
+        # Number of cases we want to fix
+        qs = CourseInstructorGrade.objects\
+            .exclude(course=None)\
```
alex-shen1, line 20: What kinds of CourseInstructorGrade instances don't have `course`?

tcf_website/management/commands/fix_multiple_objects_returned.py
```diff
@@ -0,0 +1,113 @@
...
+            .values('course', 'first_name', 'last_name')\
+            .annotate(num=Count('id'))\
+            .filter(num__gt=1)
+        if self.verbosity > 0:
+            print(qs.count())
```
alex-shen1, line 25: Can you add a description of what this is to the log i.e. `print(f'Number of duplicates: {qs.count()})`? Doubt it matters too much considering this is a rare one-off thing, but I think it's good practice to have complete logging.

tcf_website/management/commands/fix_multiple_objects_returned.py
```diff
@@ -0,0 +1,113 @@
...
+
+        qs_without_num = qs.values('course', 'first_name', 'last_name')
+        total = CourseInstructorGrade.objects\
+            .filter(reduce(or_, [Q(**a) for a in qs_without_num]))\
+            .order_by('first_name', 'last_name', 'subdepartment', 'number')
+        without_email = total.filter(email='')
+
+        # Fix the weird case with two different emails for the same professor
+        to_fix = list(CourseInstructorGrade.objects.filter(
```
alex-shen1, line 34: Could probably have a better variable name... but the best one I can think of is `conflicting_emails` which is both long and not particularly either.

tcf_website/management/commands/fix_multiple_objects_returned.py
```diff
@@ -0,0 +1,113 @@
...
+            email='lr4zs@virginia.edu',
+        ))
+        for obj in to_fix:
+            obj.email = ''
+        CourseInstructorGrade.objects.bulk_update(to_fix, ['email'])
+
+        # Now everything should have been fixed.
```
alex-shen1, line 41: What is the `Now everything should have been fixed.` comment referring to? As far as I can tell, nothing happens to `without_email` inbetween its creation and this logging message. Does `bulk_update` change `without_email`?

tcf_website/management/commands/fix_multiple_objects_returned.py
```diff
@@ -0,0 +1,113 @@
...
+        if self.verbosity > 0:
+            print(without_email.count())
```
alex-shen1, line 43: Aforementioned request to add log details

tcf_website/management/commands/fix_multiple_objects_returned.py
```diff
@@ -0,0 +1,113 @@
...
+class Command(BaseCommand):
+    """This fixes existing `MultipleObjectsReturned` errors"""
+
+    help = 'Fix `MultipleObjectsReturned` errors that were caused by inconsistent email data'
+
+    def handle(self, *args, **options):
+        self.verbosity = options['verbosity']
+
+        # Number of cases we want to fix
+        qs = CourseInstructorGrade.objects\
```
alex-shen1, line 19: I assume `qs` stands for QuerySet, in which case we might want to change this variable name to something like `erroneous_instances` that conveys more information as to what it is.

tcf_website/management/commands/fix_multiple_objects_returned.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Now everything should have been fixed.
+        if self.verbosity > 0:
+            print(without_email.count())
+
+        for a in qs_without_num:
+            instructors = list(
+                CourseInstructorGrade.objects.filter(**a).order_by('email')
+            )
+            # The second one has the email, and the first one doesn't
```
alex-shen1, line 49: Is this assuming that there are only two duplicates for any given instance? Is that a safe assumption?

tcf_website/management/commands/fix_multiple_objects_returned.py
```diff
@@ -0,0 +1,113 @@
...
+                )
+            ).values(*fields_all)  # to exclude actual_enrolled
+        )
+        # Drop problematic instances
+        instances_to_fix.delete()
+        # Create fixed instances
+        objects_created = CourseInstructorGrade.objects.bulk_create(
+            [CourseInstructorGrade(**x) for x in combined])
+        if self.verbosity > 0:
+            print(len(objects_created))
```
alex-shen1, line 96: Same log detail comment

tcf_website/management/commands/fix_multiple_objects_returned.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Check the number of `MultipleObjectsReturned` error again
+        assert CourseInstructorGrade.objects\
```
alex-shen1, line 99: Nice job writing tests!

tcf_website/management/commands/load_grades.py
```diff
@@ -67,14 +67,19 @@ def clean(self, df):
         CourseGrade.objects.all().delete()
         CourseInstructorGrade.objects.all().delete()
 
-        return df.dropna(
+        # Filter out data with no grades
+        df = df.dropna(
```
alex-shen1, line 71: Not really part of this PR per se, but same thing with `qs` as before; honestly not sure what general practice is, but I personally think we should avoid variable names that are basically just letters.

alex-shen1 (commented):
Good work! Works fine on my local machine, so most of my comments are just minor nitpicks.

tcf_website/management/commands/fix_multiple_objects_returned.py
Bartleby2718: I couldn't find a pattern, and that's out of the scope of this PR, but I'm suspecting the courses were left out when we ran the `april7.sql` three times, just like [this issue](https://discord.com/channels/752270432329269328/767189787131117609/783027475084869686). Luckily, the number wasn't too big iirc.

tcf_website/management/commands/fix_multiple_objects_returned.py
Bartleby2718: Done in f45bd0586dca0e7cfc008ddf272e194f660bbc75

tcf_website/management/commands/fix_multiple_objects_returned.py
Bartleby2718: Done in f45bd0586dca0e7cfc008ddf272e194f660bbc75

tcf_website/management/commands/fix_multiple_objects_returned.py
Bartleby2718: Clarified through comments in f45bd0586dca0e7cfc008ddf272e194f660bbc75

tcf_website/management/commands/fix_multiple_objects_returned.py
Bartleby2718: Done in f45bd0586dca0e7cfc008ddf272e194f660bbc75

tcf_website/management/commands/fix_multiple_objects_returned.py
Bartleby2718: Done in f45bd0586dca0e7cfc008ddf272e194f660bbc75

tcf_website/management/commands/fix_multiple_objects_returned.py
Bartleby2718: Clarified through comments in f45bd0586dca0e7cfc008ddf272e194f660bbc75 and added an assert statement

tcf_website/management/commands/fix_multiple_objects_returned.py
Bartleby2718: Done in f45bd0586dca0e7cfc008ddf272e194f660bbc75

tcf_website/management/commands/load_grades.py
Bartleby2718: I believe using `df` for any pandas `DataFrame` instance is very commonplace, and this is a function with very limited scope, so I don't think there's value in using a descriptive name for this parameter.

db/add_missing_instructor_emails.sql
Bartleby2718: Just verified that 1292 `Instructor` instances are updated in both cases (when you run all the commands in this PR), so I took out the unnecessary assumption in cb31ecc5c58c5de5da9715e6019dbb1a00e221f6.

alex-shen1 (approved):

PR #307, opened 2020-12-05
Add Google Tag Manager

julesfll:
# Status
Done

# What I did
Google offers two services for tracking tags: gtag.js and Google Tag Manager. GTM is a web client for no-code tag managment. I was working on gtag.js previously but decided to scrap it because I think GTM will work better for the following reasons:
- It's complicated to integrate the JS for tags when most functionality is embedded in Django views
- It will be accessible to Marketing
- It prevents the need for adding a lot of code the codebase

Sometime later I can do a brief workshop on how to use GTM
## Issues addressed
- Added boilerplate for GTM setup
- Added tags for user login, going to review page, and creating a review (submitting the form)
## Screenshots
image


j-alicia-long (approved):
Looks awesome to me! Good thinking :)

Let's merge this in soon so we can get it out for v1.1.0

PR #312, opened 2020-12-06
Login modal

### LINKED ISSUE #132: Login doesn't consider `?next=`
# Problem
If you are not logged-in and try to create a new review, it doesn't respect the `?next=/reviews/new` path. Therefore, whenever you login, it always sends you to `/accounts/profile`, which is the default `LOGIN_REDIRECT_URL` of the social-auth pipeline. This is not what we want. Instead, we want to send the user to where they were trying to go! 

### Note 
If this problem is not fixed, it will become a problem later when we make users login before seeing reviews, as they will be sent to `/accounts/profile` instead of the same page.

# Solution
I'm going to work on the Authentication pipeline to make a user's current location relevant factor into the redirect URL upon login.

### LINKED ISSUE #272: Login overhaul
_(Still open to discussion.)_
### Option A
Login page takes you back to previous page you were looking at
### Option B
Login replaced with modal/popup instead of special login page

j-alicia-long:
# Status
Done

# What I did
- Updated UI for Login 
     - Created login modal so users aren't forced to be redirected
     - Redirect to browse page with error if login fails
- Linked modal on "Sign In" links on navbar, sidebar, landing page

## Issues addressed
- This closes #132 
- This closes #272 

## Screenshots
| Login modal      | Login Error |
| ----------- | ----------- |
|  image | image |


tcf_website/templates/about/logo-tagline.html
```diff
@@ -0,0 +1,10 @@
+<!-- Logo with name & tagline, import where needed -->
```
julesfll, line 1: Good refactor

tcf_website/templates/login/login_modal.html
```diff
@@ -0,0 +1,27 @@
...
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title">Sign In</h5>
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+          <div class="d-flex justify-content-center">
```
julesfll, line 13: Would this be a place to use your `logo-tagline`?

julesfll (commented):
Modal looks good, and this is a good simplification! I do have some suggestions though for usability.
- If you log in incorrectly the first time, you get redirected to `/login/error`. Then if you log in correctly, instead of getting redirected to `Browse`, you get redirected to `/login/error` again (and the message is still there), so it looks you logged in incorrectly again
- You can still log in when you're already logged in, should probably be disabled (or force redirect when logged in to browse)
- If you log in incorrectly from your current page, you lose where you were by being redirected to `login`

What would solve all these issuse would be to eliminate the login page entirely, which might be the best move. The messages communicate a successful/unsuccessful login, so you can just redirect to the current page no matter the outcome.

tcf_website/templates/login/login_modal.html
j-alicia-long: I tried but it looks kinda bad in the modal on small screens rip

j-alicia-long:
> Modal looks good, and this is a good simplification! I do have some suggestions though for usability.
> 
> * If you log in incorrectly the first time, you get redirected to `/login/error`. Then if you log in correctly, instead of getting redirected to `Browse`, you get redirected to `/login/error` again (and the message is still there), so it looks you logged in incorrectly again
> * You can still log in when you're already logged in, should probably be disabled (or force redirect when logged in to browse)
> * If you log in incorrectly from your current page, you lose where you were by being redirected to `login`
> 
> What would solve all these issuse would be to eliminate the login page entirely, which might be the best move. The messages communicate a successful/unsuccessful login, so you can just redirect to the current page no matter the outcome.

Good point! I'll look into removing the page. Thought my design of it was pretty ugly anyway ngl

tcf_website/templates/login/login_modal.html
julesfll: Ok maybe something to look into later
We need to put that refactor to use 😄

j-alicia-long:
> Modal looks good, and this is a good simplification! I do have some suggestions though for usability.
> 
> 1. If you log in incorrectly the first time, you get redirected to `/login/error`. Then if you log in correctly, instead of getting redirected to `Browse`, you get redirected to `/login/error` again (and the message is still there), so it looks you logged in incorrectly again
> 2. You can still log in when you're already logged in, should probably be disabled (or force redirect when logged in to browse)
> 3. If you log in incorrectly from your current page, you lose where you were by being redirected to `login`
> 
> What would solve all these issuse would be to eliminate the login page entirely, which might be the best move. The messages communicate a successful/unsuccessful login, so you can just redirect to the current page no matter the outcome.

So removing the login page turned out to be more tricky than anticipated with django auth. I basically replaced the login page in the workflow with the browse page. 
- Fixed the first issue with an if-statement checking the view for including the "next" redirect.
- Fixed the second issue with a force redirect to profile/browse at the `/login` url
- Couldn't fix third issue due to aforementioned trickiness

We might need an overhaul of the auth system in general since its inflexibility has been causing other issues with extra data collection.

tcf_website/views/auth.py
```diff
@@ -6,19 +6,22 @@
...
 
 def login(request):
     """Login view."""
     if request.user.is_authenticated:
-        return redirect('browse')
-    return render(request, 'login/login.html')
-
+        messages.success(request, "Logged in successfully!")
+        return redirect('profile')
+    return browse(request)
```
alex-shen1, line 16: Just curious — when does this return get reached?

tcf_website/views/auth.py
```diff
@@ -53,4 +56,4 @@ def logout(request):
     """Logs out user."""
     auth_logout(request)
     messages.add_message(request, messages.SUCCESS, "Logged out successfully!")
-    return redirect('login')
+    return redirect('browse')
```
alex-shen1, line 59: What's the difference between `browse(request)` like the others have and this?

tcf_website/views/auth.py
j-alicia-long: When the user isn't authenticated and they try to go to `/login`

tcf_website/views/auth.py
j-alicia-long: I can't use `redirect('browse')` in `login` or `login_error` views without causing a `ERR_TOO_MANY_REDIRECTS`. I'm honestly not sure entirely why

j-alicia-long:
@julesfll I opened Login Error Redirect Issue #328 to address the third comment!

PR #325, opened 2020-12-14
Review form fixes

### LINKED ISSUE #317: Review form still enables dropdown prematurely
It's possible to open the next dropdown before the API fills it with options. This isn't fantastic, but isn't terribly important as you have to click REALLY fast to see the empty dropdown before it fills.

Definitely an async issue... whoever figures out a similar issue with #155 should tackle this too; might be me (Alex) but might also not be

j-alicia-long:
# Status
Done

# What I did
It's a minor fix for a edge case bug that @alex-shen1 found that I can't reproduce (clicking on a subsequent dropdown before API call finishes will show no options). I'm putting the dropdown enable code in the callback to fix that. 

Also tacked on some minor UI updates
- Write a Review page: list last name first for professor dropdown so you can type more easily
       - Instructor API: Ordering by last name 
- My reviews page: add a new review button if you still have reviews so you can write many in a row

## Issues addressed
This closes #317 hopefully

## Screenshots
Before
image

New review button
image


tcf_website/static/reviews/new_review.js
```diff
@@ -83,18 +86,19 @@ jQuery(function($) {
             $.each(data.results, function(i, instr) {
                 $("<option />", {
                     val: instr.id,
-                    text: instr.first_name + " " + instr.last_name
+                    text: instr.last_name + ", " + instr.first_name
```
jackliu612, line 89: I agree last name first makes searching easier, but are we concerned at all that this is the only place on the site with this format (as far as I can tell)?

jackliu612 (commented):
Overall it looks good to me. Pulled the branch and it looks good at all window sizes. I can't reproduce the bug mentioned but the fix is good nonetheless.

tcf_website/static/reviews/new_review.js
j-alicia-long: This is how it was on the old site so I think it should be alright

Plus the new instructor dropdown also has the astname-firstname format

tcf_website/static/reviews/new_review.js
```diff
@@ -114,10 +118,11 @@ jQuery(function($) {
                 }).appendTo("#semester");
             });
             return this;
-        });
-
-        // Enable semester selector, disable the following
-        $("#semester").prop("disabled", false);
+        })
+            .done(function() {
+                // Enable semester selector, disable the following
```
alex-shen1, line 123:
```suggestion
                // Enable semester selector
```
Nothing following the semester selector

alex-shen1 (approved):

Issue #299, opened 2020-12-03
Add PYTHONUNBUFFERED to Dockerfile


Bartleby2718:
- There was a comment in `Dockerfile` that I took out in 84c5af53e711b8d1d883e283069958f5c17e66d9, and that is `# ENV PYTHONUNBUFFERED 1`. It was a comment anyway, so I thought there was no harm in taking it out, but it seems to make a difference when`-i` flag is not used to run a `docker` command.
- Reference: https://stackoverflow.com/questions/59812009/what-is-the-use-of-pythonunbuffered-in-docker-file

- Without `ENV PYTHONUNBUFFERED=1`
```
$ docker exec tcf_django python3 manage.py load_grades ALL_DANGEROUS 
100%|██████████| 1802/1802 [00:01<00:00, 1683.52it/s]
100%|██████████| 1898/1898 [00:00<00:00, 2078.78it/s]
100%|██████████| 2090/2090 [00:00<00:00, 2207.40it/s]
100%|██████████| 2113/2113 [00:01<00:00, 2089.40it/s]
100%|██████████| 1980/1980 [00:00<00:00, 1980.75it/s]
100%|██████████| 2050/2050 [00:01<00:00, 2002.17it/s]
100%|██████████| 1946/1946 [00:01<00:00, 1884.29it/s]
100%|██████████| 2284/2284 [00:01<00:00, 1982.09it/s]
100%|██████████| 2188/2188 [00:00<00:00, 2247.34it/s]
100%|██████████| 2739/2739 [00:01<00:00, 2051.63it/s]
100%|██████████| 2438/2438 [00:01<00:00, 2050.88it/s]
100%|██████████| 2473/2473 [00:01<00:00, 2200.82it/s]
100%|██████████| 2544/2544 [00:01<00:00, 2231.07it/s]
100%|██████████| 2667/2667 [00:01<00:00, 1939.60it/s]
100%|██████████| 2602/2602 [00:01<00:00, 2141.84it/s]
100%|██████████| 2628/2628 [00:01<00:00, 1895.30it/s]
100%|██████████| 6411/6411 [00:00<00:00, 11762.38it/s]
100%|██████████| 14116/14116 [00:01<00:00, 8623.39it/s]
Fetch Course and Instructor data for later use
Found 46852 sections in 2011_fall.csv
Done with loading CSV files
Found 49348 sections in 2012_fall.csv
Done with loading CSV files
Found 54340 sections in 2012_spring.csv
[...]

PR #331, opened 2020-12-14
Env vars

### LINKED ISSUE #329: Use a different settings.py for each environment
- The current `settings.py` depends heavily on a single environment variable called `DEBUG`, which not only makes the `settings.py` very cluttered but also makes it impossible for us to simulate `DEBUG=True` in the local development environment. This is especially a problem when we want to test `403.html`, `404.html`, or `500.html`
- Reference: https://code.djangoproject.com/wiki/SplitSettings#SimplePackageOrganizationforEnvironments
- Notable changes
  - [x] Update `BASE_DIR`
  - [x] Remove some code that had been commented out (e.g. `STATIC_URL`, `HOSTNAME`, `PUBLIC_IPV4`, `ALLOWED_HOSTS`)
  - [x] Move `django_heroku.settings(locals())` to the bottom (hence removed `LOGGING`)
  - [x] Run silk in only the development environment
- A possible side effect 
  - [x] Remove `UserWarning` from `django-environ`?
- Minor style changes in `base.py`
  - [x] Add a trailing comma in `INSTALLED_APPS`
  - [x] Replace double quotes with single quotes

Bartleby2718:
# Status
- In Progress (**DON'T MERGE THIS!**)
- If this looks good to both @alex-shen1 and @jackliu612, I will finish deployment through GitHub Actions, reset all secrets (because dumb me leaked them all), and then ditch TravisCI once and for all
- I expect to see some errors in _some_ environment (either Heroku or production) at some point, although I hope not.

# What I did
- Changed how `settings.py` is structured
- Used environment variables more extensively
- See #329 and [this Google Drive folder](https://drive.google.com/drive/u/1/folders/10_L1AbgHBpO6eaelbJg10Ar2xz_l0P3L) for detail

## Issues addressed
Closes #329


.github/workflows/python-app.yml
```diff
@@ -28,11 +28,10 @@ jobs:
...
         env:
-          POSTGRES_USER: tcf_django
-          POSTGRES_PASSWORD: s3kr1t
-          POSTGRES_DB: tcf_db
-        ports:
-          - 5432:5432
+          POSTGRES_USER: ${{ secrets.DB_USER }}
+          POSTGRES_PASSWORD: ${{ secrets.DB_PASSWORD }}
+          POSTGRES_DB: ${{ secrets.DB_NAME }}
+          POSTGRES_PORT: ${{ secrets.DB_PORT }}
```
alex-shen1, line 34: Github won't let me check what the values of the repo secrets are; are these parameters for connecting to prod DB or establishing a local test DB? Is it advisable to be running tests against the prod DB?

.github/workflows/python-app.yml
```diff
@@ -51,7 +50,7 @@ jobs:
 
       - name: Run pylint
         run: |
-          pylint --load-plugins pylint_django tcf_website tcf_core
+          pylint --disable=R0801 --load-plugins pylint_django tcf_website tcf_core
```
alex-shen1, line 53: Would it be better to individually have `pylint disable` lines in the specific files where this is an issue? I think there's also a way to have pylint config files, but not sure if that's worth the extra effort.

tcf_core/settings/base.py
```diff
@@ -1,6 +1,6 @@
-# pylint: disable=line-too-long,import-error,fixme
+# pylint: disable=fixme
 """
-Django settings for tcf_core project.
+Base Django settings for tcf_core project.
 
 Generated by 'django-admin startproject' using Django 3.0.4.
```
alex-shen1, line 5: Not very relevant, but should we remove this? lol

tcf_core/settings/base.py
```diff
@@ -12,25 +12,27 @@
...
 
 import os
+
 from django.urls import reverse_lazy
 from django.contrib.messages import constants as messages
 import environ
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
alex-shen1, line 21: What does this do?

tcf_core/settings/prod.py
```diff
@@ -0,0 +1,16 @@
...
+
+# The following if statement is needed to prevent overwriting global variables when
+# this settings file is interpreted
+if os.environ.get('DJANGO_SETTINGS_MODULE') == 'tcf_core.settings.prod':
+    # SECURITY WARNING: App Engine's security features ensure that it is safe to
+    # have ALLOWED_HOSTS = ['*'] when the app is deployed. If you deploy a Django
+    # app not on App Engine, make sure to set an appropriate host here.
+    # See https://docs.djangoproject.com/en/dev/ref/settings/ (from GCP
+    # documentation)
+    ALLOWED_HOSTS = ['*']
```
alex-shen1, line 13: I don't really know why, but when I was trying to do manual GAE deployment when Travis was down, I had to add `thecourseforum.com` to `ALLOWED_HOSTS` or the deployment wouldn't work (see 40cac033ca14b5c379e5845f9c3870605cdac62d). Might be better to keep it in; is that a security problem?

tcf_core/settings/testing.py
```diff
@@ -0,0 +1,18 @@
+# pylint: disable=import-error,unused-wildcard-import,wildcard-import
+"""Django settings module for testing environment such as Heroku"""
```
alex-shen1, line 2: IMO the name of this file should be something like `dev`, because at least to me "testing" connotes something along the lines of written test cases or a scratch file where some various settings were experimented with, which this is not.

tcf_core/settings/ci.py
```diff
@@ -0,0 +1,13 @@
...
+Django settings module for CI services such as Travis or GitHub Actions
+
+Unlike other files where we use environment variables for everything, it is
+okay to use some strings constants here and in .travis.yml or python-app.yml
+"""
+from .base import *
+
+# The following if statement is needed to prevent overwriting global variables when
+# this settings file is interpreted
+if os.environ.get('DJANGO_SETTINGS_MODULE') == 'tcf_core.settings.ci':
```
alex-shen1, line 12: Don't forget to add these variables to GitHub Secrets when we switch over to GA!

tcf_core/settings/base.py
```diff
@@ -121,20 +123,20 @@
...
+STATIC_ROOT = os.path.join(BASE_DIR, 'static/')
 
 
 # Database
 # https://docs.djangoproject.com/en/3.0/ref/settings/#databases
 
 DATABASES = {
     'default': {
-        'NAME': 'tcf_db',
+        'NAME': env.str('DB_NAME'),
```
alex-shen1, line 134: Just writing this as a reminder: we'll have to update `.env` in the tCF drive once this gets merged to avoid breaking things.

tcf_core/settings/base.py
```diff
@@ -174,80 +170,30 @@
...
-        'NAME': env.str('DB_NAME'),
-        'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'USER': env.str('DB_USER'),
-        'PASSWORD': env.str('DB_PASSWORD'),
-        'HOST': env.str('DB_HOST'),
-        'PORT': env.str('DB_PORT'),
-        'OPTIONS': {'sslmode': 'require'},
-    }
-
-LOGGING = {
```
alex-shen1, line 224: Are you sure about just deleting logging like this?

tcf_core/settings/base.py
```diff
@@ -174,80 +170,30 @@
...
-    'version': 1,
-    'disable_existing_loggers': False,
-    'handlers': {
-        'console': {
-            'class': 'logging.StreamHandler',
-        },
-    },
-    'root': {
-        'handlers': ['console'],
-        'level': 'WARNING',
```
alex-shen1, line 234: As a side note, maybe we want log levels to go up to `DEBUG`? Not sure if we actually have logs at that level though.

.env.example
```diff
@@ -1,22 +1,31 @@
...
-SECRET_KEY=
-SOCIAL_AUTH_GOOGLE_OAUTH2_KEY=
-SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET=
+# Django
+DJANGO_SETTINGS_MODULE=$DJANGO_SETTINGS_MODULE
+DEBUG=$DEBUG
+SECRET_KEY=$SECRET_KEY
 
-# Dev and Prod settings
-ES_PUBLIC_API_KEY=search-7gvgdz431agb5agermmcuvx8
```
alex-shen1, line 8: rip looks like exporting the branch zip also exported commit history

.env.example
```diff
@@ -1,22 +1,31 @@
...
+ES_INSTRUCTOR_SEARCH_ENDPOINT=$ES_INSTRUCTOR_SEARCH_ENDPOINT
+ES_COURSE_DOCUMENTS_ENDPOINT=$ES_COURSE_DOCUMENTS_ENDPOINT
+ES_INSTRUCTOR_DOCUMENTS_ENDPOINT=$ES_INSTRUCTOR_DOCUMENTS_ENDPOINT
+ES_PRIVATE_API_KEY=$ES_PRIVATE_API_KEY
+
+# discord
+DISCORD_URL_BUG=$DISCORD_URL_BUG
+DISCORD_URL_FEEDBACK=$DISCORD_URL_FEEDBACK
+
+# No longer used but
```
alex-shen1, line 29: Is it safe to just delete these?

tcf_core/settings/prod.py
alex-shen1: Even if CI works for the foreseeable future, for future debugging purposes we should probably have a comment saying that adding `thecourseforum.com` is necessary for manual deployment.

alex-shen1 (changes requested):
Haven't gotten the chance to personally test w/ different `.env`s, but it looks pretty good! I'm not sure that we need to store the local DB credentials as env vars when they're basically arbitrary text, but the code is significantly cleaner now because of it.

Again, don't forget to update the technical docs in the drive accordingly!

.github/workflows/python-app.yml
Bartleby2718: These are the env vars in `.env.ci`, which you can find in the Google Drige folder I mentioned in the description.

tcf_core/settings/base.py
Bartleby2718: `__file__` is a special variable for the current file (i.e. `tcf_core/settings/base.py` in this case). Calling `.dirname()` on the absolute path to this file will return the path to the parent directory (i.e. `tcf_core\settings`). If we do that two more times (so 3 in total), we can the path to the root directory. The original code needed only two `.dirname()` because `settings.py` was in `tcf_core`, but we need one more because the depth increased by 1.

.env.example
Bartleby2718: If you look at the code, they're not used anywhere because they're all commented out. Normally I would delete these, but I was reminded of `# PYTHONUNBUFFERED = 1` in `Dockerfile`, which I removed because it was a comment but turned out to be helpful. (That's #299 btw.)

tcf_core/settings/base.py
Bartleby2718: We've always debugged issues by reproducing them on Heroku/local and seeing the logs there (rather than on GAE), but `django_heroku` configures logging, as mentioned in the description.

tcf_core/settings/base.py
Bartleby2718: I actually looked into the source code of `django_heroku`, and it is set to DEBUG: https://github.com/heroku/django-heroku/blob/8396b5544c36274942ee6d16ab46d4cddcdd3b4b/django_heroku/core.py#L117.

tcf_core/settings/base.py
alex-shen1: I thought `django_heroku` was just for Heroku - does it set up GAE logging? GAE logs (or rather the error logs, which I'm not sure is the same thing as this?) have been pretty helpful, so I think we should keep those if we're not already.

Bartleby2718: Oh were they helpful? I thought no one was using it. I'll put it back.

tcf_core/settings/base.py
Bartleby2718: Done in 255a6169bf8ac84c20edbe8c929d8a7248700b29.

.github/workflows/python-app.yml
Bartleby2718: I did this because it's a cross-file check, but now I don't see the error because of the changes I made afterwards, so I removed the flag entirely in d8436a63dc705ee2b4aa07444dfb3adc975dacea.

tcf_core/settings/base.py
Bartleby2718: Removed in ece568ea729e7072ed9a42f67061f81745158677 and now the file is 11 lines shorter.

tcf_core/settings/prod.py
Bartleby2718: Done in 6b694b981a082fcf033da727468b94e5b71baa7d.

tcf_core/settings/testing.py
Bartleby2718: I still think it should be `testing` because we use Heroku to test major functionalities (possibly with the help of D&M) before deploying to production.
https://en.m.wikipedia.org/wiki/Deployment_environment#Environments.

tcf_core/settings/base.py
Bartleby2718: Done in 255a6169bf8ac84c20edbe8c929d8a7248700b29.

Bartleby2718:
Superceded by #344

PR #333, opened 2020-12-15
Add a user-friendly Course endpoint

### LINKED ISSUE #332: Add a user-friendly Course endpoint
- Users must click a few times or search for class code to get to `/course/7610` , the course page for CS 3240, even though they know exactly where they want to go.
- Since students are already familiar with such mnemonic/number pairs, why don't we just let them use them (as in https://vagrades.com/uva/CS3240)?
- We can potentially extend this to other models like department (although dealing with space or special characters will require some discussion).

Bartleby2718:
# Status
Done

# What I did
- Add a user-friendly Course endpoint 
- See #332 for the rationale

## Need to discuss
- Should I redirect to the normal URL (`/course/{course_id}` or just keep the URL? I think redirecting to the original URL partly defeats the purpose of this PR.
  - If we do keep the URL, should I allow or disallow mixed/lower case mnemonics? I believe we should allow them because some keyboards on mobile default to `Title Case`.
- Should I refrain from accessing the database twice? I think the overhead is very minimal in this case, so it should acceptable (for readability).

## Issues addressed
- Closes #332

## Screenshots
image
image
Set `DEBUG = False` in the very bottom of `settings.py` to see the 404 page.

## Note
- We badly need a 404 page .


julesfll:
> We badly need a 404 page

I can work on that today

j-alicia-long (commented):
Code looks good! to reply to some of your discussion points:

1. I agree that if we update this, we should default to using the new one and remove the old ones.
     - One concern: If you remove the original endpoint, we'd have to update Professor Bloomfield ahead of time to update the way Lou's List links our pages. Maybe best to do this after course registration week? 
     - Alternatively, have the old endpoint redirect to the new one for now (in case ppl have them bookmarked).
2. I think you should allow mixed/lower case mnemonics but just convert/redirect them to the lowercase version.
3. Is there any notable difference in django-silk?

Btw, have you looked into using Django's built-in SlugField for this? https://learndjango.com/tutorials/django-slug-tutorial

Bartleby2718:
@j-alicia-long Hmm, do you think I should change `course_instructor` as well?

j-alicia-long:
> @j-alicia-long Hmm, do you think I should change `course_instructor` as well?

Sure, are names unique though? is there a way to generate unique slugs for them?

Bartleby2718:
@j-alicia-long It clearly is possible--we can use the ID or add a random string--but I'm not sure if instructor slug will be as useful as course mnemonic+number. Maybe we can do it at another time?

tcf_website/tests/test_browse.py
```diff
@@ -0,0 +1,31 @@
+# pylint: disable=no-member
+"""Tests for browse.py."""
+from django.test import TestCase
+from django.urls import reverse
+
+from .test_utils import setup
+
+
+class CourseViewTestCase(TestCase):
```
j-alicia-long, line 9: Thanks for writing tests!

tcf_website/views/browse.py
```diff
@@ -70,17 +70,30 @@ def department(request, dept_id):
...
+    course = get_object_or_404(Course, pk=course_id)
+    return redirect('course',
+                    mnemonic=course.subdepartment.mnemonic,
+                    course_number=course.number)
 
-    course = Course.objects.get(pk=course_id)
+
+def course_view(request, mnemonic, course_number):
+    """A new Course view that allows you to input mnemonic and number instead."""
+    # Redirect if the mnemonic is not all uppercase
```
j-alicia-long, line 83: Are urls inherently case-sensitive? hmm

j-alicia-long (approved):
I tested it a bit and the redirect works great! Maybe we could write a Selenium script later to verify uniqueness. Good job!

tcf_website/views/browse.py
Bartleby2718: Apparently, [yes](https://stackoverflow.com/a/7996997).

PR #335, opened 2020-12-16
Login Fix

vn6:
# Status
Done

# What I did
- fixed bug with extra info form the first time a user logs in

## Issues addressed
n/a

## Screenshots
n/a


j-alicia-long (approved):
Oh bless! thanks for looking into this

PR #341, opened 2020-12-19
Refactor and fix course_instructor page

### LINKED ISSUE #338: 500 error when (course, instructor) pair does not exist
Should be 404

### LINKED ISSUE #339: GPA is shown as 0.0 for pass/fail courses in some places
image
image

### LINKED ISSUE #340: "Other" is not shown in the pie chart
After some research, I have come to a conclusion that "Other" means "Pass" in a CR/NC grading scheme. (I have no idea how it's gonna change in semesters with GC, though.)

Bartleby2718:
# Status
Done

# What I did
- Refactor `course_instructor` in `browse.py` and everything in `course_data.js`
- Minor bug fixes
- Small performance improvement

## Issues addressed
- This PR fixes #338, fixes #339, and fixes #340

## Screenshots
1. Before the PR
image
2. After the PR
image
Kimberly and I chose this position/color: `#6775A6`.
3. Performance
image
The OS/Reiss page loads 50 ms faster (based on the average of middle three requests).


tcf_website/views/browse.py
```diff
@@ -157,74 +154,57 @@ def course_instructor(request, course_id, instructor_id):
...
+        )
+    data = {key: safe_round(value) for key, value in data.items()}
 
     try:
         grades_data = CourseInstructorGrade.objects.get(
             instructor=instructor, course=course)
-
+    except ObjectDoesNotExist:  # if no data found
+        pass
+    # NOTE: Don't catch MultipleObjectsReturned because we want to be notified
```
alex-shen1, line 184: Is the idea that the site should just crash if this is the case? Is there a better way of logging this?

tcf_website/static/course/javascript/course_data.js
```diff
@@ -1,63 +1,97 @@
 const loadData = data => {
-    const { a_plus, a, a_minus, b_plus, b, b_minus, c_plus, c, c_minus, d_plus, d, d_minus, f, withdraw, drop } = data;
-    const grades_data = [a_plus, a, a_minus, b_plus, b, b_minus, c_plus, c, c_minus, d_plus, d, d_minus, f, withdraw, drop];
+    const { a_plus, a, a_minus, b_plus, b, b_minus, c_plus, c, c_minus, d_plus, d, d_minus, f, ot, drop, withdraw } = data;
```
alex-shen1, line 2: Considering that more students likely drop than withdraw, I think the graphs look better with the "last" slice of orange being thicker, as opposed to having a thin slice:

image

Would love to check how the old tCF did it...but the old site is now down.

alex-shen1: Also, I think the `ot` variable should be renamed to something like `other` (or maybe `pass`, as that's what it seems like it represents most of the time?) for clarity - it was a little ambiguous to me what it meant until I read further into the PR.

alex-shen1 (changes requested):
The refactors look excellent; just have a few small style comments. Also not a huge fan of the olive green color, but I don't have any good recommendations off the top of my head. My instinct is to just use one of the blues, although I think we had a discussion about this before and it didn't look great. Maybe someone else (@j-alicia-long? @julesfll?) can take a look at it?

tcf_website/views/browse.py
Bartleby2718: I believe the standard approach is to [send emails to devs when there are server errors](https://docs.djangoproject.com/en/3.1/howto/error-reporting/#server-errors). Sendgrid is a popular service in the field, and it seems to allow [100 emails/day for free forever](https://sendgrid.com/pricing/). The free plan isn't so scalable but is probably sufficient for our use case, and this can be done in a separate PR.

tcf_website/static/course/javascript/course_data.js
Bartleby2718: 1. `withdraw` wasn't there before, and I added that. I thought it was also a noteworthy piece of information, but let me know if you think I should remove it since I added it without asking the team. If not, I'll just change the order.
2. Yeah, I'll rename it. I just kept the way it is, but renaming it will definitely increase readability/maintainability. Also, the data files containing `OT` also uses `DR` for drop and `W` for withdraw, which means previous maintainers didn't always keep the original names in the data files.

Let me know whether I should discard `withdraw`, and I'll make the above changes in a single commit.

Bartleby2718: Edit: Sorry, it seems like `withdraw` was already there and I only changed the order of  `drop` and `withdraw`.

Bartleby2718: Renamed all instances of `ot` to `other` except when it's used for the name of the model (646a83b6593ab5a173f1512fdd39087731a0cfe8 and 12d305d05916269a7c47012416b85dc077d5c2f9). I'm not 100% sure, but I think changing a model field *might* break the database because we create migration files on the fly every time.

Bartleby2718: Fixed in 97690bd601faf21beff4bdfc9ff3ff2fcbabc1d0 (color: `#6775A6`)

alex-shen1 (approved):
Sorry for taking so long to approve - looks good!

PR #348, opened 2020-12-20
Added better feedback for review account wall

j-alicia-long:
# Status
Done? Need input.

# What I did
Fixed issue with `new_review` url redirecting back to browse when user is not authenticated
- Clicking on Review sidebar link will pop open login modal
- Trying to navigate to `/reviews/new` url will redirect to browse and show an error alert "Login is required to view this page"

## Additional Notes
Actually not the best solution, but it'll do for now. 
- If we want a fully fluid login modal (that doesn't redirect to browse on login error) we likely will have to create a custom AJAX `login_required` decorator https://stackoverflow.com/questions/312925/django-authentication-and-ajax-urls-that-require-login
- Not sure this is worth doing until we make a decision on the login system overhaul, possibly to add MS exchange/professor accounts

## Screenshots
| Sidebar link click     | URL navigation |
| ----------- | ----------- |
| image | image |

### DIFF
--- a/tcf_website/templates/base/sidebar.html
+++ b/tcf_website/templates/base/sidebar.html
@@ -13,7 +13,9 @@
          <hr>
       </a>
     </li>
+    <!-- Account wall for writing reviews -->
     <li>
+    {% if user.is_authenticated %}
       <a href="{% url 'new_review' %}"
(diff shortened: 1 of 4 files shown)


Bartleby2718:
Superseded by #354 but forgot to close

PR #354, opened 2021-01-14
Set LOGIN_URL

### LINKED ISSUE #349: /reviews/new URL is bad if attempt made at manually accessing it
Not really _too_ much of a problem because I don't really see any user trying to access the URL manually on prod, but if they do, then this appears:

image

Key issues just being the **Sign In** button be considered "active" and the URL technically pointing to login when it's just the normal browse page (I believe the view specifically just passes the return of the `browse` view if some condition is met).

Bartleby2718:
# Status
Done

# What I did
- Set `LOGIN_URL` in `settings.py`

## Issues addressed
- Supersedes PR #348
- Closes #349 "/reviews/new URL is bad if attempt made at manually accessing it"

## Screenshots
Users are immediately redirected to the Google Sign In page when accessing a page that requires login (e.g. `/reviews/new` or `/profile`)
image
After that, they see the original page they wanted to access.


alex-shen1 (commented):
Works fine for me! At least, what I tested. I do have one question though - I'm not quite sure how the URL routing works, so I'm not sure if this would work for the review autofill feature (put a button on course_instructor pages that automatically fills the review form w/ relevant data) that's still in development. You can check out the code [here](https://github.com/thecourseforum/theCourseForum2/compare/addreview_on_reviews#diff-f612fa96516131a4fd19aabbf9cd3a8726f21655c824ba2adf36336db1a7da45R168), but in essence the current structure is that the template will submit a (hidden) form via POST request to the review page, which grabs the data off the request. If all the redirecting does is simply save the URL and then restore it once the user logs in, then we might run into some issues? If you have any insight into this, that'd be good to know.

Bartleby2718:
@alex-shen1 I glanced over it, and it seems like the client is sending an additional HTTP POST request to get the data needed for autopopulation. I guess this will also work, but when I had thought about this, I considered using query parameters instead of a separate POST request. That way, it matches the semantics of the HTTP methods--GET is a [safe](https://developer.mozilla.org/en-US/docs/Glossary/Safe) method while POST isn't--and does not require another HTTP request (which translates to load time).

Is there anything else you'd like to do before merging this PR?

alex-shen1:
I considered using GET as well, but I thought that adding all those parameters to the URL would look kinda ugly in the browser, so I thought we should hide them with POST requests. Can you elaborate on what you meant by "query parameters" (if you meant something other than just adding `?`s to the GET requests)? Just for further consideration later; assuming that this PR won't be the point of failure in the future, I'm gonna go ahead and approve.

alex-shen1 (approved):

Bartleby2718:
@alex-shen1 That is exactly what I meant, and that is also why I didn't work on that issue. I agree that the URL will become ugly. Sorry I can't offer any alternatives.

alex-shen1:
RE: safeness, I'm not too familiar with the concept, but I'm fairly certain that the POST request doesn't modify any data on the site (it's just a glorified form submission), so it shouldn't be too much of an issue? I haven't worked on the feature in a while, but I'll finish it off at some point; we can pick up this discussion more then.

Issue #274, opened 2020-11-24
Pretty 404 page


julesfll:
Bump of #134 

The current 404 page looks like this:
https://thecourseforum.com/this-page-does-not-exist
![image](https://user-images.githubusercontent.com/4724192/101143774-5f71ab80-35e5-11eb-9e1d-0828c2813941.png)

To do:
- Create 404/403/500 pages that looks better and provides more information to the user.
- Extend from `base.html` so navbar/sidebar are still present

j-alicia-long:
Wait @julesfll were you working on this?

PR #355, opened 2021-01-14
Improve review form

### LINKED ISSUE #248: Write review form validation
Add robust backend validation to review form

### LINKED ISSUE #266: Edit reviews
- User should be able to edit reviews they have written
- Edit icon should show on `My Reviews` and possibly anywhere on the site that the review shows up

Has been worked on already (stale)
 https://github.com/thecourseforum/theCourseForum2/tree/edit-review

Bartleby2718:
# Status
Done

# What I did

## Issues addressed
Closes #248 "Write review form validation"
Closes #266 "Edit reviews"

## Screenshots
image

## Notes
1. Not sure if I should rename `new_review.css` and `.new-review` class.
2. Need to delete related branches once merged (`edit-reviews`, `review-modelform`, `edit-review`)


tcf_website/static/reviews/sync_review_sliders.js
```diff
@@ -0,0 +1,15 @@
+/* Sync Review Rating Slider InputIds */
+jQuery(($) => {
+    const tuples = [
+        ["#instructor_rating", "#instructor_rating2"],
+        ["#enjoyability", "#enjoyability2"],
+        ["#difficulty", "#difficulty2"],
+        ["#recommendability", "#recommendability2"]
+    ];
+
+    tuples.forEach((tuple) => {
```
j-alicia-long, line 10: Nice refactor, but perhaps add some in-text comments since it's a little less clear what the code is doing now

tcf_website/views/review.py
```diff
@@ -52,61 +59,18 @@ def new_review(request):
         # TODO: use a proper django form.
```
j-alicia-long, line 59: Nice! Can this comment be removed now?

tcf_website/views/review.py
```diff
@@ -123,3 +87,23 @@ def get_object(self):  # pylint: disable=arguments-differ
...
+
+
+@login_required
+def edit_review(request, review_id):
+    """Review modification view."""
+    review = get_object_or_404(Review, pk=review_id)
+    if review.user != request.user:
+        raise PermissionDenied('You are not allowed to edit this review!')
+
+    if request.method == 'POST':
```
j-alicia-long, line 99: Does this set the `modified` field of the `Review` model automatically?

Btw any insights on using `default=timezone.now` vs `auto_now` in the model? I think `auto_now` will update it automatically on `Model.save()` but not sure about the former.
https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.DateField.auto_now

j-alicia-long (commented):
Great work, esp with the refactoring! Glanced over the code and made a couple of comments.

Sidenote: This would be a great feature to write integration tests (or end-to-end? UI?) for if we want to start looking into that.

tcf_website/views/review.py
Bartleby2718: Yup! Just did.

tcf_website/static/reviews/sync_review_sliders.js
Bartleby2718: Added a couple lines of comments.

tcf_website/views/review.py
Bartleby2718: Good catch! I don't think using the `default` parameter will set `modified` upon modification, so changed those.

Bartleby2718:
That would be ideal, but I don't think we can log in via NetBadge using Selenium. 🤔

tcf_website/templates/reviews/edit_review.html
```diff
@@ -0,0 +1,30 @@
...
+
+{% block title %}Edit a Review - theCourseForum{% endblock %}
+
+{% block styles %}
+<link rel="stylesheet" href="{% static 'reviews/new_review.css' %}" />
+{% endblock %}
+
+{% block content %}
+<div class="new-review container text-left mx-auto">
+  <h1 class="h2 mt-2 mb-4">Edit a review</h2>
```
alex-shen1, line 12: Not sure if this is significant, but there's an `<h1>` opening tag and an `</h2>` closing tag here.

tcf_website/views/review.py
```diff
@@ -123,3 +86,23 @@ def get_object(self):  # pylint: disable=arguments-differ
...
                 "You are not allowed to delete this review!")
         return obj
+
+
+@login_required
+def edit_review(request, review_id):
+    """Review modification view."""
+    review = get_object_or_404(Review, pk=review_id)
+    if review.user != request.user:
+        raise PermissionDenied('You are not allowed to edit this review!')
```
alex-shen1, line 96: Attempting to edit a review you don't own by manually changing the URL results in this error screen:

image

This isn't terribly important because the key thing is that the permission checking works and it's unlikely anyone would try, but is there a easy way to more elegantly show this error?

tcf_website/views/review.py
```diff
@@ -123,3 +86,23 @@ def get_object(self):  # pylint: disable=arguments-differ
...
+
+    if request.method == 'POST':
+        form = ReviewForm(request.POST, instance=review)
+        if form.is_valid():
+            form.save()
+            messages.success(request,
+                             f'Successfully updated the review for {form.instance.course}!')
```
alex-shen1, line 103: Maybe should be "updated **your** review"?

alex-shen1 (changes requested):
Good work! Outside of these minor nitpicks, the functionality works fine on my machine and I can't get it to break. 

RE: testing, do we strictly need Selenium? Do you think there's any value in creating tests along the lines of just creating dummy Review instances locally, calling the functions directly in a test file, and seeing if they work as expected?

tcf_website/views/review.py
Bartleby2718: Fixed in cbd75728f4dfd61dde70533958f9143247dc6ead

tcf_website/views/review.py
Bartleby2718: This should be done as part of #274:
> Create 404/403/500 pages that looks better and provides more information to the user.

Typical end users won't try that, so I believe this is outside the scope of this PR.

tcf_website/templates/reviews/edit_review.html
Bartleby2718: Fixed in ff3e5e44011622b06df1adf3827d358a091d404a

tcf_website/tests/test_review.py
```diff
@@ -0,0 +1,77 @@
...
+            'difficulty': self.review1.difficulty,
+            'recommendability': self.review1.recommendability,
+            'enjoyability': self.review1.enjoyability,
+            'amount_reading': self.review1.amount_reading,
+            'amount_writing': self.review1.amount_writing,
+            'amount_group': self.review1.amount_group,
+            'amount_homework': self.review1.amount_homework,
+        })
+
+    def test_nonexistent_review_id(self):
```
alex-shen1, line 32: Seems like this test and the one after it prints unnecessary error messages to console:

image

This issue seems similar to what @joek13 is dealing with when testing the delete review functionality (#359) - maybe when one of you guys figures it out you should let the other know.

tcf_website/templates/reviews/edit_review.html
```diff
@@ -9,7 +9,7 @@
 
 {% block content %}
 <div class="new-review container text-left mx-auto">
-  <h1 class="h2 mt-2 mb-4">Edit a review</h2>
+  <h1 class="h2 mt-2 mb-4">Edit a review</h1>
```
alex-shen1, line 12: I'm pretty sure it's already been established why we need `<h1>` tags with `<h2>` classes, but I just find it funny

alex-shen1 (changes requested):
Tests pass on my machine, which is good! Definitely should work out some way to suppress the error logs though.

tcf_website/tests/test_review.py
Bartleby2718: Fixed in 027149ed53a72a0c7ba41eac567c1ab6ddf5f7d1

alex-shen1 (approved):
Nice job! I think everything has been addressed at this point, so I'm gonna hit approve.

PR #359, opened 2021-01-23
Add 'deleted review' message on reviews page

### LINKED ISSUE #357: Add "deleted review" message on reviews page
Editing/creating a new review will create banner-style messages on the review page that look like this:

image
image

However, no such message currently exists for when a review is deleted. Simply add a banner in the same style, alerting the user that their review has been deleted.

joek13:
## Status
- Done

- This PR closes #357 

## What I did
- Inspected code that adds message for newly created reviews in `views.new_review`
- Added override for generic `DeleteReview`'s `delete()` method. The override adds a message to the request, then invokes superclass `delete()` so that the review is properly deleted

## Screenshots
- Before: deleting a review leaves no message or confirmation
Screen Shot 2021-01-22 at 8 09 12 PM
- After: deleting a review leaves a banner-style message confirming successful deletion
Screen Shot 2021-01-22 at 8 09 30 PM

## Tests
Steps to test:
1. Create a new review
2. From the reviews page, delete your newly created review
3. Observe a message indicating your review has been deleted

- It would probably be most idiomatic to use [SuccessMessageMixin](https://docs.djangoproject.com/en/3.1/ref/contrib/messages/#django.contrib.messages.views.SuccessMessageMixin) from Django. I tried this, but [this StackOverflow issue](https://stackoverflow.com/questions/24822509/success-message-in-deleteview-not-shown) explains why that doesn't work.
- Is it appropriate code style to link to a StackOverflow issue in my code? It helps contextualize the choice I made, but it might also be a little clunky...

## Merging the PR
- Who should merge this PR?
  - [ ] I will merge it myself
    - [ ] preserve the history
    - [x] squash-merge

### DIFF
--- a/tcf_website/tests/test_review.py
+++ b/tcf_website/tests/test_review.py
@@ -7,7 +7,6 @@ from django.contrib.messages import get_messages
 from django.test import TestCase
 from django.urls import reverse
 
-from .test_utils import setup
 from .test_utils import setup, suppress_request_warnings
 
 
@@ -80,6 +79,7 @@ class EditReviewTests(TestCase):
         messages = [m.message for m in get_messages(response.wsgi_request)]
         self.assertIn('difficulty', str(messages[0]))
 
+
 class DeleteReviewTests(TestCase):
     """Tests for the DeleteReview view."""
 
@@ -101,6 +101,7 @@ class DeleteReviewTests(TestCase):
         messages = [m.message for m in get_messages(response.wsgi_request)]
         self.assertEqual(str(messages[0]), 'Successfully deleted your review!')
 
+    @suppress_request_warnings
     def test_delete_nonexistent_review_id(self):
         """Test if a 404 is returned for deleting a nonexistent review ID."""
         self.client.force_login(self.user1)
@@ -110,6 +111,7 @@ class DeleteReviewTests(TestCase):
 
         self.assertEqual(response.status_code, 404)
 
+    @suppress_request_warnings
     def test_unauthorized_user_delete(self):
         """Test if a 403 is returned for unauthorized review deletion."""
         self.client.force_login(self.user2)  # force login as user2
--- a/tcf_website/tests/test_review.py
+++ b/tcf_website/tests/test_review.py
@@ -1,0 +1,57 @@
+# pylint: disable=no-member
+"""Tests for Review model."""
+
+from django.contrib.messages import get_messages
+from django.test import TestCase
+from django.urls import reverse
+
+
+from .test_utils import setup
+
+
+class DeleteReviewTests(TestCase):
+    """Tests for the DeleteReview view."""
+
+    def setUp(self):
+        setup(self)  # set up tests: add some example data
+
+    def test_delete_review_message(self):
+        """Test if a message is shown when a user deletes their review."""
+        self.client.force_login(
+            self.review1.user)  # force a login as the author of review1
+        # try and make the user delete review1
+        response = self.client.post(
+            reverse('delete_review', args=[self.review1.id]),
+        )
+
+        self.assertEqual(response.status_code, 302)  # ensure 302 Found status
+
+        # get messages from the request
+        messages = [m.message for m in get_messages(response.wsgi_request)]
+        self.assertEqual(str(messages[0]), 'Successfully deleted your review!')
+
+    def test_delete_nonexistent_review_id(self):
+        """Test if a 404 is returned for deleting a nonexistent review ID."""
+        self.client.force_login(self.user1)
+        response = self.client.post(
+            reverse('delete_review', args=[0])  # id 0 = nonexistent review
+        )
+
+        self.assertEqual(response.status_code, 404)
+
+    def test_unauthorized_user_delete(self):
+        """Test if a 403 is returned for unauthorized review deletion."""
+        self.client.force_login(self.user2)  # force login as user2
+
+        # Note: because PermissionDenied is (should be) raised in this view, we get a really ugly
+        # traceback in the test runner. This isn't so nice, because it makes the
+        # `./precommit` output annoying to sort through.
+        # Is there any way to silence the errors while we're running tests?
+        response = self.client.post(
+            # try to delete review1, which is not authored by user2
+            reverse('delete_review', args=[self.review1.pk])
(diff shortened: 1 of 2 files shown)


tcf_website/views/review.py
```diff
@@ -114,7 +114,6 @@ class DeleteReview(LoginRequiredMixin, generic.DeleteView):
     """Review deletion view."""
     model = Review
     success_url = reverse_lazy('reviews')
-    success_message = "Successfully deleted the review for %(course)s"
```
joek13: Sorry I missed this in the first commit. This is a vestige from when I tried to implement the message using `SuccessMessageMixin`, but I ended up overriding `delete` instead.

j-alicia-long:
Great work! Glad you were able to follow the docs and set up your local project so quickly :)

I'll give this a test and a code review tomorrow!

tcf_website/views/review.py
```diff
@@ -123,3 +123,19 @@ def get_object(self):  # pylint: disable=arguments-differ
             raise PermissionDenied(
                 "You are not allowed to delete this review!")
         return obj
+
+    def delete(self, request, *args, **kwargs):
+        """Overide DeleteView's delete function to add a message confirming deletion."""
+        # Note: we don't use SuccessMessageMixin because it currently has issues
+        # with DeleteViews.
+        # See:
```
alex-shen1, line 131: To answer your question regarding SO, there are several other places in the codebase where people have linked to SO posts, so I think it should be fine. IMO, more comments are almost always better than less (as it stands currently, tCF definitely needs more documentation...).

However, maybe move the `See:` to line 130 (which would move the link to 131) so we don't have that long stretch of white space.

tcf_website/views/review.py
```diff
@@ -123,3 +123,19 @@ def get_object(self):  # pylint: disable=arguments-differ
...
+        # https://stackoverflow.com/questions/24822509/success-message-in-deleteview-not-shown
+        course = super().get_object().course
+
+        messages.add_message(
+            request,
+            messages.SUCCESS,
+            'Successfully deleted the review for ' +
```
alex-shen1, line 138: I think this should be `deleted your review` - what do you think?

alex-shen1 (changes requested):
Works as expected for me! Nice job getting set up and completing your first feature so quickly. Just a few minor nitpicks, and then it'll be good to merge.

tcf_website/views/review.py
joek13: Agreed. Will do.

tcf_website/views/review.py
joek13: That's more concise and gets the message across better. I originally had something like this, but I used the "Successfully deleted the review for ..." phrasing to be consistent with your screenshots in #357, which both follow the format "Successfully created/updated/deleted the review for <class>."

That said, I prefer the more concise version and will implement that.

joek13:
I've added some tests for the `DeleteReview` view. They check to make sure my message is displayed, but I also added tests for deleting a nonexistent review and trying to delete someone else's review. My tests are based extensively on PR #355—thanks for the help there.

I tried to follow the `test_review.py` from PR #355 as closely as possible to minimize headaches when we have the inevitable merge conflict. (Is this the proper way to do it? Should I have done something else?)

tcf_website/views/review.py
alex-shen1: Whoops, I meant `deleted your review for (course)`, as I think stating the course information to confirm is important to avoid having users worry about potentially deleting the wrong review. Regarding the screenshots in the issue I made, those were taken from a PR (#355) where I left the same feedback...my bad, should've made that clear. Sorry!

tcf_website/tests/test_review.py
```diff
@@ -0,0 +1,57 @@
+# pylint: disable=no-member
+"""Tests for Review model."""
```
alex-shen1, line 2: Running the tests locally gives me this:

image

It looks like tests are passing, but as you can see there's quite a bit of junk that's also printed that I don't think we want to be there...

tcf_website/tests/test_review.py
```diff
@@ -0,0 +1,57 @@
...
+
+        self.assertEqual(response.status_code, 404)
+
+    def test_unauthorized_user_delete(self):
+        """Test if a 403 is returned for unauthorized review deletion."""
+        self.client.force_login(self.user2)  # force login as user2
+
+        # Note: because PermissionDenied is (should be) raised in this view, we get a really ugly
+        # traceback in the test runner. This isn't so nice, because it makes the
+        # `./precommit` output annoying to sort through.
```
alex-shen1, line 48: bless you for actually remembering to use `precommit` lmao

alex-shen1: On a more serious note, I spent a bit of time trying to figure out a way to solve the problem, but the best I got was setting `sys.tracebacklimit=0`, which gets rid of most of the stack trace, yielding this:

image

Also, saw some [discussion on StackOverflow](https://stackoverflow.com/questions/8391411/how-to-block-calls-to-print) about blocking print statements entirely by redirecting `sys.stdout`, but when I tried it, the errors still printed out all the same.

The reduction is still an improvement, but not enough IMO. This is definitely something we want to figure out before merging into `dev`.

alex-shen1 (changes requested):
Nice job writing all these tests! Testing is definitely something we're trying to ramp up going forward, so this is good. Just need to address the overprinting issue you're already familiar with and it should be good to go.

alex-shen1:
Looks like @Bartleby2718 figured out a way to do it with 027149ed53a72a0c7ba41eac567c1ab6ddf5f7d1 - you should be able to just follow his work here once it gets merged into `dev`.

joek13:
Eeek... sorry about the failed builds. I fixed the merge conflict in the web editor (instead of locally), and it looks like `pylint` has detected some sins. I'm fixing now and running `./precommit`.

alex-shen1 (approved):
Looks good to me! Congrats on finishing your first feature!

alex-shen1:
As a side note, I don't think you really have to worry about intermediate builds failing - as long as the build is passing by the time someone gets around to reviewing it, it's all good.

Issue #319, opened 2020-12-11
Add text on empty Course page


j-alicia-long:
add some text on empty semester course page https://thecourseforum.com/course/6580
e.g. Looks like this course isn't taught this semester. Sort by "All" in the top right to see previous semesters

PR #374, opened 2021-02-06
Course page empty

quyenxhuynh:
## Status
- Done

## GitHub Issues addressed (`#319`)
- This PR closes 319

## What I did
- Added an additional context variable named `this_semester_instructors` in `views/browse.py` to check if there are any instructors/classes teaching this semester. 
- If there is not, show message in `templates/course/course.html`
- Added `.new` and `.new.hide` classes to `common/show-hide.css`, `common/show-hide.js` so it message disappears when user clicks "All"

## Screenshots
- Before

- After

## Tests
- `./course/CLAS/2010` - Shows card
- `./course/CLAS/2010` > Click "All" - Card should disappear
- `./course/CLAS/2020` - Should not show the card
- `./course/CLAS/2020` > Click "All" - Should not show the card

N/A

## Merging the PR
- Who should merge this PR?
  - [ ] I will merge it myself
  - [ ] preserve the history
  - [x] squash-merge


tcf_core/settings/base.py
```diff
@@ -7,7 +7,10 @@
 import environ
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+BASE_DIR = os.path.dirname(
+    os.path.dirname(
+        os.path.dirname(
+            os.path.abspath(__file__))))
```
n4mahajan, line 13: Seems like there might have been some inadvertent reformatting issues? This happened to me too when I ran precommit but I'm not too sure what the solution is. From what I've been told, I think we're supposed to discard such changes if possible (which is pretty easy to do with Github Desktop if you don't already use it).

tcf_website/templates/course/course.html
```diff
@@ -34,6 +34,11 @@ <h2>{{ course.title }}</h2>
             <br/>
             <br/>
 
+            <!-- Unsure if we want to remove it when they are looking at all semesters. If so, may need the opposite of "old" class. -->
+            {% if not this_semester_instructors %}
+                <h5 class="">Looks like this course isn't taught this semester. Sort by "All" in the top right to see previous semesters.</div>
```
n4mahajan, line 39: Might I recommend rephrasing this to "Looks like this course isn't being taught this semester."?

j-alicia-long (commented):
Good job! Definitely something we were missing.

Recommend putting the text in a white card/div just so it's consistent with our empty reviews display:
image

Example: https://thecourseforum.com/course/7189/5835

tcf_website/views/browse.py
```diff
@@ -123,7 +126,8 @@ def course_view(request, mnemonic, course_number):
                       'course': course,
                       'instructors': instructors,
                       'latest_semester': latest_semester,
-                      'breadcrumbs': breadcrumbs
+                      'breadcrumbs': breadcrumbs,
+                      'this_semester_instructors': this_semester_instructors
```
alex-shen1, line 130: Seems like all you're using this variable for in the template is to check whether or not it exists — might be better to just have a Boolean like `taught_this_semester`. This is mainly just for readability purposes, as you'd just end up moving the nullchecking from the template to the view.

alex-shen1 (changes requested):
Looks good! Just have one concern that shouldn't take too long to wrap up. As a side note, you might want to revert the random autoformatting changes to make the PR cleaner (sorry! we really have to figure out what's going on with that lol)

tcf_website/templates/course/course.html
```diff
@@ -34,6 +34,19 @@ <h2>{{ course.title }}</h2>
             <br/>
             <br/>
 
+            {% if not taught_this_semester %}
```
Bartleby2718, line 37: Can't we use this instead?
https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#for-empty

Bartleby2718 (dismissed):

tcf_website/templates/course/course.html
quyenxhuynh: I don't think so, since these queries include previous semesters (we just hide them if they're in the current semester tab)

Bartleby2718: Ahh you're right.

tcf_website/views/browse.py
```diff
@@ -104,6 +104,9 @@ def course_view(request, mnemonic, course_number):
             semester_last_taught=Max('section__semester',
                                      filter=Q(section__course=course)),
         )
+    taught_this_semester = Instructor.objects.filter(
+        section__course=course, section__semester=latest_semester).distinct().exists()
```
Bartleby2718, line 108: We can take out `.distinct()`

Bartleby2718, line 108:
```suggestion
    taught_this_semester = Section.objects.filter(course=course, semester=latest_semester).exists()
```

j-alicia-long (approved):
Looks good to me!

alex-shen1 (approved):
Dunno why it's saying needs review when @j-alicia-long already approved, so maybe tacking on approval from me will fix it.

PR #399, opened 2021-02-14
Prevent user from spamming submit for reviews/multiple reviews for sa…

n4mahajan:
## Status
- Done unless further modifications are needed

- This PR closes 

## What I did
- If required fields are filled (i.e. Subject, Course, Instructor, Semester, and description if user writes anything), removes submit button after being clicked so that user can't spam click it or click it multiple times. Avoids multiple submissions of the same review.
- Also created backend validation when creating a new review so that a user doesn't create a review for the same course under the same instructor/semester again. Instead gives them an error message to edit their original review. Avoids multiple reviews from same user and further ensures multiple submissions aren't sent.

## Screenshots
NOTE: these screenshots are outdated
- Before
Screen Shot 2021-02-14 at 1 26 21 AM
Screen Shot 2021-02-14 at 5 42 17 PM
Screen Shot 2021-02-14 at 5 45 00 PM

- After
Screen Shot 2021-02-14 at 5 42 49 PM

^The two above reviews were permitted in "before". Once I tried to recreate the review again for the same course with the same professor/semester, it gives an error message now. I can add the course name and such to the error message though?

Screen Shot 2021-02-14 at 5 45 41 PM

^Submit button is removed if review is successfully filled and submitted

## Tests
- Go to the review page to write a review. Click the submit button without filling anything to ensure the button doesn't disappear/nothing on the page is modified. 
- Now fill out the required fields (description if you want to) for any random course. Click the submit button and ensure it disappears so that you can't click it (you can try "spamming" what was there before). This should successfully submit the review.
- Go back to write another review. Select the same course/instructor/semester that you did before. Try submitting the review, and ensure that instead of creating and saving your review in your review page, it instead gives you an error message to edit your original review. 
- Go back to write another review once again. Select the same course that you did before, but with a different professor. You should be able to successfully submit this review. Likewise, write another review with the same professor as the original review, but a different semester. This should also be successful. The only time a duplicate review is not allowed is when it is the same course for the same instructor AND semester.

- I tried simply timing out the button and disabling the button once clicked, but doing this makes the error/required messages not pop if fields are not filled out, and also doesn't submit the review. To bypass this, I opted to simply timeout the button and remove/hide it for 2 seconds, if and only if it is clicked and all the required fields are filled (i.e. the review is being successfully submitted).
- For backend validation, I didn't know if there's any unique ID or code for a given course for a particular instructor and semester. I know there is a general course ID, but this applies to any instructor/semester for the course. So instead, I checked if the user had already created a review for a given course under the same instructor AND semester before by using the Review model's Semester field and __str__.

## Merging the PR
- Who should merge this PR?
  - [ ] I will merge it myself
  - [x] preserve the history
  - [ ] squash-merge


j-alicia-long:
Will give this a review tmrw but this is really useful! tyty for the extra validation

tcf_website/views/review.py
```diff
@@ -63,6 +63,17 @@ def new_review(request):
             instance.hours_per_week = \
                 instance.amount_reading + instance.amount_writing + \
                 instance.amount_group + instance.amount_homework
+
+            # Verify that user has not already submitted a review for this
+            # specific course for the same semester and instructor before
+            for review in request.user.reviews():
+                if review.__str__() == instance.__str__() and review.semester == instance.semester:
```
alex-shen1, line 70: This doesn't seem like a very good way to compare Reviews... for example, I can review CS 2110 twice as long as I put a different instructor. 

image

I would do something like `review.course == instance.course` — disclaimer: haven't tested that, so that might not work. The idea is that we should be comparing data directly instead of a string representation.

tcf_website/views/review.py
```diff
@@ -63,6 +63,17 @@ def new_review(request):
...
+                    messages.error(
+                        request,
+                        'Looks like you have already submitted a review for this specific course!'
+                        ' Please edit your original review if you would like to modify anything.')
```
alex-shen1, line 74: I can imagine that someone might be upset if (for whatever reason) they forgot they reviewed a course already, typed out a long, thoughtful, comment, and hit submit, only for this message to pop up and their work is gone. 

What do you think about modifying the "Submit Review" button to first show a popup modal presenting them with the option of going back to the review form or just overwriting their old review?

tcf_website/templates/reviews/new_review.html
```diff
@@ -31,5 +31,27 @@ <h1 class="h2 mt-2 mb-4">Write a review</h1>
...
+
+  // Checks if required fields for review were filled
+  // If required fields are filled, timeout and hide the submit button so that the user can't spam click it
+  $(document).ready(function () {
+    $("#submitBtn").click(function () {
+        var reviewForm = document.getElementById("reviewForm");
+        // Check if required fields are filled
+        if (reviewForm.checkValidity()) {
+            var btn = document.getElementById("submitBtn");
+            btn.style.display = 'none';
```
alex-shen1, line 43: What do you think about simply disabling the button instead of hiding it entirely? Removing the element shifts the whole page, which is a little jarring, and I don't think I've ever seen something like this elsewhere. IMO an even better approach might just be to have some backend validation instead of disabling the frontend, but I don't want to invalidate the work you've done so far.

alex-shen1 (changes requested):

tcf_website/views/review.py
alex-shen1: One exception I can think of: during Covid, some profs have been co-teaching and alternating lectures (Wu/Brunelle for Algo, Sherriff/McBurney for 3240), so in that sense people _might_ want to review separately? But I highly doubt that people would go to the extent to specifically do this.

tcf_website/templates/reviews/new_review.html
n4mahajan: Agreed, disabling the button is definitely better. However, when I tried doing that, it also disabled form submission and disabled the required field messages, so that if you clicked submit and didn’t fill out the required fields, nothing would happen and it wouldn’t tell you why. I can take out the frontend validation and simply opt for backend validation though.

tcf_website/views/review.py
n4mahajan: Yeah I considered being able to review for the exact same course but different instructor and thought that was generally fine, in case you interacted with that professor in some other capacity especially now with co-teaching as you mentioned. I didn’t think writing reviews for identical courses but different instructors was problematic so I left it out but I’m open to changing this. I doubt people would bother repeating reviews for different instructors anyway.

n4mahajan: Will test your suggestion though along with other methods though

tcf_website/views/review.py
n4mahajan: I definitely thought about that and I think you’re definitely right. I thought it was a somewhat fringe case but it should certainly still be accounted for so I’m glad you mentioned it. I didn’t really know how to do it but I’ll look into since it’s probably worth it.

tcf_website/views/review.py
Bartleby2718: Barring Alex's corner cases, I would use something like this (might not be accurate Django code):
```python
reviews_on_same_class = request.user.review_set.filter(
    course=instance.course,
    instructor=instance.instructor,
    semester=instance.semester,
)
if reviews_on_same_class.exists():
    ...
```

n4mahajan: As of right now I've made it specific to the same course + instructor + semester (as @Bartleby2718 suggested), but I can change it to prevent reviewing the overall course again? (i.e. with any instructor/semester; so only filter for course=instance.course)

tcf_website/templates/reviews/new_review.html
```diff
@@ -31,5 +31,27 @@ <h1 class="h2 mt-2 mb-4">Write a review</h1>
   $(function () {
     $('[data-toggle="tooltip"]').tooltip()
   });
+
+  // Checks if required fields for review were filled
+  // If required fields are filled, timeout and hide the submit button so that the user can't spam click it
+  $(document).ready(function () {
+    $("#submitBtn").click(function () {
```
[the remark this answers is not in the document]
n4mahajan: Agreed, but I sometimes had to switch like with "document.getElementById("reviewForm").submit();" because it wouldn't work with $("#...") for some reason when I tried?

tcf_website/templates/reviews/new_review.html
```diff
@@ -26,32 +28,44 @@ <h1 class="h2 mt-2 mb-4">Write a review</h1>
...
+    // Stop normal submit to check for duplicate review with ajax request
+    $("#reviewForm").submit(function(e) {
+        e.preventDefault();
 
-    function removeButton() {
-        //sets button back to initial state
-        btn.style.display = 'initial';
-        document.getElementById("submitBtn").disabled = false;
-    }
+        $.ajax({
```
alex-shen1, line 43: Quite the clever solution to what we were talking about!

tcf_website/templates/reviews/duplicate_review_modal.html
```diff
@@ -0,0 +1,20 @@
...
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+          It looks like you've already submitted a review for this class! If you would like, you can go to your
+          original review to edit it as you see fit.
+      </div>
+      <div class="modal-footer">
+        <button type="button" class="btn btn-warning" data-dismiss="modal">Close</button>
```
alex-shen1, line 16: I don't think the yellow looks great considering the style guide...maybe change it to orange like the submit button?

(for any other reviewers, this is what it looks like)
image

tcf_website/templates/reviews/new_review.html
```diff
@@ -26,32 +28,44 @@ <h1 class="h2 mt-2 mb-4">Write a review</h1>
 {% block js %}
 <script type="text/javascript" src="{% static 'reviews/new_review.js' %}"></script>
 <script type="text/javascript" src="{% static 'reviews/sync_review_sliders.js' %}"></script>
+<script src="jquery-3.5.1.min.js"></script>
```
alex-shen1, line 31: We already have jquery linked in `index.html` and it gets inherited, so this isn't necessary.

tcf_website/views/review.py
```diff
@@ -83,7 +73,32 @@ def new_review(request):
     return render(request, 'reviews/new_review.html')
 
 
+@login_required()
+def check_duplicate(request):
+    """Check for duplicate reviews when a user submits a review
+    based on if it's the same course with the same instructor/semester.
+    Used for an ajax request in new_review.html"""
```
alex-shen1, line 80: Super pedantic, but I believe Ajax is a proper noun?

tcf_website/views/review.py
```diff
@@ -83,7 +73,32 @@ def new_review(request):
     return render(request, 'reviews/new_review.html')
 
 
+@login_required()
+def check_duplicate(request):
```
alex-shen1, line 77: This is something that doesn't matter for prod, but if I log out and manually to to `http://localhost:8000/reviews/check_duplicate`, it first prompts me to log in (which is good) but then crashes with this error:

image

Realistically this isn't something that we should ever have to worry about, but it's probably good to just add some basic error checking to catch this.

tcf_website/views/review.py
```diff
@@ -83,7 +73,32 @@ def new_review(request):
...
+    based on if it's the same course with the same instructor/semester.
+    Used for an ajax request in new_review.html"""
+
+    form = ReviewForm(request.POST)
+    instance = form.save(commit=False)
+
+    reviews_on_same_class = request.user.review_set.filter(
+        course=instance.course,
+        instructor=instance.instructor,
+        semester=instance.semester
```
alex-shen1, line 88: I still think at the very _least_ we shouldn't allow submitting for the same class/instructor pair on different semesters. I think there's also a compelling case for removing instructors from this filter as well, but that's at least debatable. @Bartleby2718 my interpretation of the example code snippet you posted was just a format to do the filtering instead of taking a stance on this matter; if this wasn't the case, can you clarify?

tcf_website/views/review.py
```diff
@@ -83,7 +73,32 @@ def new_review(request):
...
+    # Review already exists so it's a duplicate; inform user
+    if reviews_on_same_class.exists():
+        response = {"duplicate": True}
+        return JsonResponse(response)
+
+    response = {"duplicate": False}
+    return JsonResponse(response)
+
 # Note: Class-based views can't use the @login_required decorator
+
```
alex-shen1, line 100: What's with these extra whitespace lines? I think this comment is supposed to be directly above `DeleteReview`?

alex-shen1 (changes requested):
Good stuff -  I like your idea of using Ajax to trigger a modal. Just a few minor points from me.

tcf_website/views/review.py
Bartleby2718: I don't have a strong opinion on this. I was just suggesting what could be potentially more readable code.

tcf_website/templates/reviews/duplicate_review_modal.html
n4mahajan: Agreed, should def fit the site theme and I made it rather quickly so I'll update it.

alex-shen1 (approved):
Good stuff!

alex-shen1 (approved):

PR #418, opened 2021-02-28
Disable django-silk on Heroku/GAE

### LINKED ISSUE #411: Remove django-silk from production
## Describe the bug
`django-silk` shouldn't be on production, but it is.

## To Reproduce
Related tables can be found in the production database...

## Expected behavior
`django-silk` should be installed and active only on local environment.

## Screenshots
N/A

## Your Environment:
 - OS: N/A
 - Browser: N/A

## Additional context
This is on me because I misunderstood how Python files were interpreted.

Bartleby2718:
## Status
- Done

- This PR closes #411.

## What I did
- Moved silk-related code inside an `if` statement so that it doesn't get executed on Heroku or GAE.

## Screenshots
| Before | After |
|:---:|:---:|
|image | image |

## Tests
1. Run `docker exec -ti tcf_db psql tcf_db -U tcf_django` and then run `\dt` inside the PostgreSQL shell to verify that silk-related tables are there, just like in the screenshot.
2. Change the `if` statement to `if False:`
3. Run `docker exec -ti tcf_db psql tcf_db -U tcf_django` and then run `\dt` inside the PostgreSQL shell to verify that silk-related tables are gone. From this, we see that the `if` statement does something.
4. Now see `tcf_core/settings/prod.py`. Since we were able to modify `ALLOWED_HOSTS` the same way on GAE, we see that this PR will fix the issue.

- After this PR is merged to `master`, we should delete/truncate silk-related tables on prod to reduce the size of database dump. Right now, I uploaded a dump without some silk-related entries, which is 10 times smaller than the original.

## Merging the PR
- Who should merge this PR?
  - [ ] I will merge it myself
    - [ ] preserve the history
    - [x] squash-merge

### DIFF
--- a/tcf_core/settings/__init__.py
+++ b/tcf_core/settings/__init__.py
@@ -17,9 +17,9 @@ def custom_recording_logic(request):
     return not request.path.startswith('/api')
 
 
+# Performance profiling for non-API views during development
 if os.environ.get('DJANGO_SETTINGS_MODULE') not in [
         'tcf_core.settings.dev', 'tcf_core.settings.prod']:
-    # Performance profiling for non-API views during development
     INSTALLED_APPS.append('silk')
     MIDDLEWARE.append('silk.middleware.SilkyMiddleware')
     SILKY_INTERCEPT_FUNC = custom_recording_logic
--- a/tcf_core/settings/__init__.py
+++ b/tcf_core/settings/__init__.py
@@ -17,7 +17,9 @@ def custom_recording_logic(request):
     return not request.path.startswith('/api')
 
 
-# Performance profiling for non-API views during development
-INSTALLED_APPS.append('silk')
-MIDDLEWARE.append('silk.middleware.SilkyMiddleware')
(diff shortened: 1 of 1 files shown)


tcf_core/settings/__init__.py
```diff
@@ -17,7 +17,9 @@ def custom_recording_logic(request):
     return not request.path.startswith('/api')
 
 
-# Performance profiling for non-API views during development
-INSTALLED_APPS.append('silk')
-MIDDLEWARE.append('silk.middleware.SilkyMiddleware')
-SILKY_INTERCEPT_FUNC = custom_recording_logic
+if os.environ.get('DJANGO_SETTINGS_MODULE') not in [
+        'tcf_core.settings.dev', 'tcf_core.settings.prod']:
+    # Performance profiling for non-API views during development
```
alex-shen1, line 22: I think this comment should be above the `if` line because it describes the functionality of the entire block.

alex-shen1 (changes requested):

alex-shen1 (approved):

PR #423, opened 2021-03-05
Improve instructor_view

### LINKED ISSUE #210: Improve efficiency of instructor subdepartment query
Currently, the database is queried for every subdepartment key to find its name. @jackliu612 pointed out this was inefficient, and proposed this instead:

N+1 problem explained here as well as solutions using prefetch: https://scoutapm.com/blog/django-and-the-n1-queries-problem.

### LINKED ISSUE #422: "Server Error" shown for invalid instructor_id in instructor_view
## Describe the bug
`Server Error (500)` when an invalid `instructor_id` is used in `/instructor/{instructor_id}`

## To Reproduce
Steps to reproduce the behavior. For example:
1. Go to https://thecourseforum.com/instructor/1234234

## Expected behavior
Users should see a 404 error, not 500.

## Screenshots
image

Bartleby2718:
## Status
- Done

- This PR closes #210 and closes #422.

## What I did
- #210: Fixed a bug where 500 error was returned when a 404 should be
- #422: Improved the performance of `instructor_view`
- Wrote tests for `instructor_view`

## Screenshots
Three `Instructor`s were chosen based on the following query (`Instructor`s with most `Section`s except for `TBA`):
```
>>> from tcf_website.models import *
>>> from django.db.models import *
>>> Section.objects.values('instructors').annotate(i=Count('instructors')).order_by('-i')
```
Top row: before
Bottom row: after
image

## Tests
- Open http://localhost:8000/instructor/{instructor_id} with http://dev.thecourseforum.com/instructor/{instructor_id} side-by-side and check correctness
  - I compared a couple of them, and they were identical (except that some numbers on dev/prod ended with `.0` instead of `.00`)
- Compare the relative loading time with, say, `dev`

- @alex-shen1 told me not to delete methods that are no longer used, but I deleted `get_courses`. Let me know if I should revert that.

## Merging the PR
- Who should merge this PR?
  - [x] I will merge it myself
    - [x] preserve the history
    - [x] squash-merge


tcf_website/views/browse.py
```diff
@@ -213,67 +214,66 @@ def course_instructor(request, course_id, instructor_id):
...
+            ) / 3,
+            last_taught=Concat(
+                F('semester_last_taught__season'),
+                Value(' '),
+                F('semester_last_taught__year'),
+                output_field=CharField(),
+            ),
+    ).values('dept', *course_fields)\
+        .order_by('dept', 'name')
+    courses: Dict[str, List[Dict[str, Any]]] = {}
```
alex-shen1, line 264: So my understanding of this is that the `courses` variable is a `Dict` that has department name strings corresponding to a list of course data (each in its own `Dict`) for every course in that department, whereas `queryset` is a simpler, ungrouped list of course data. For clarity, I think it's better to rename `queryset` to `courses` and `courses` to `grouped_courses` or something like that. I'm not too committed to the name I proposed for `courses`, but I definitely think `queryset` alone doesn't convey enough information.

alex-shen1: Also, I think there should be whitespace between the last `order_by` line in the block related to `queryset` and the start of where you format `courses`, but I see that you've kept them all together so that might just be a stylistic difference.

tcf_website/views/browse.py
```diff
@@ -213,67 +214,66 @@ def course_instructor(request, course_id, instructor_id):
...
+    for course in queryset:  # type: Dict[str, Any]
+        course['avg_rating'] = safe_round(course['avg_rating'])
+        course['avg_difficulty'] = safe_round(course['avg_difficulty'])
+        course['avg_gpa'] = safe_round(course['avg_gpa'])
+        course['last_taught'] = course['last_taught'].title()
+        courses.setdefault(course['dept'], []).append(course)
+    context: Dict[str, Any] = {
```
alex-shen1, line 271: I think this should also have whitespace before it

tcf_website/tests/test_instructor.py
```diff
@@ -58,3 +59,17 @@ def test_average_difficulty_no_reviews(self):
         self.assertTrue(
             self.instructor.average_rating_for_course(self.course) is
             None)
+
+    def test_instructor_view(self):
+        """Test if context variables are correct in the instructor view."""
+        response = self.client.post(
+            reverse('instructor', args=(self.instructor.id,)))
+        self.assertEqual(response.context[0]['avg_difficulty'], 2.67)
```
alex-shen1, line 67: I'm assuming you manually calculated these expected values from the reviews in `test_utils`? This might be an irrelevant concern, but if the test instances ever changed, we'd have to edit these hardcoded values as well. It looks like other tests in the file (`test_average rating`, etc.) just manually grab the values and run an average calculation within the test, then compare with the output, so that might be the way to go.

tcf_website/views/browse.py
```diff
@@ -213,67 +214,66 @@ def course_instructor(request, course_id, instructor_id):
...
+                Avg('review__enjoyability') +
+                Avg('review__recommendability')
+            ) / 3)
+    course_fields: List[str] = ['name', 'id', 'avg_rating', 'avg_difficulty',
+                                'avg_gpa', 'last_taught']
+    queryset: List[Dict[str, Any]] = Course.objects\
+        .filter(section__instructors=instructor, number__gte=1000)\
+        .prefetch_related('review_set')\
+        .annotate(
+            dept=F('subdepartment__name'),
```
alex-shen1, line 235: Do you think it's important to make the distinction between Department and Subdepartment clear? This has been an issue for a while and it might be outside the scope of this PR, but as you know we mix them a bunch (I've been guilty of this a lot lol) and that might not be fantastic.

alex-shen1 (changes requested):
Re: some points you brought up in the PR, I think deleting `get_courses` is probably fine, and the stats ending in `.00` instead now is actually probably a good thing. Other than that, just have a few style comments and we should be good to go!

tcf_website/tests/test_instructor.py
Bartleby2718: Fixed in 9fef6f7af4ecccbf86f36aa07dd6a07cc226a80b

tcf_website/views/browse.py
Bartleby2718: I tried not to change anything in the template, but I agree that subdepartment is a better choice. The full name was not available (`ValueError: The annotation 'subdepartment' conflicts with a field on the model.`), so I used `subdept`, which is already used in `tcf_website/static/reviews/new_review.js` and `tcf_website/views/search.py`. Also fixed in 001b6592654e68ce1811a4652c6a21c4e4866f46

tcf_website/views/browse.py
Bartleby2718: Also fixed in 001b6592654e68ce1811a4652c6a21c4e4866f46

tcf_website/views/browse.py
Bartleby2718: Ah, yes. I was focusing on the functionality and neglected coming up with good variable names. Also, the `.values()` line does need whitespace before it. Fixed in 001b6592654e68ce1811a4652c6a21c4e4866f46

tcf_website/views/browse.py
alex-shen1: Maybe try `subdepartment_name` or `subdept_name`? Kinda unimportant cz it's just used within this view, though.

Bartleby2718: Renamed to `subdepartment_name` in fc412e6f3dcaedb15a25e27a359bb88af2075187.

PR #430, opened 2021-03-09
Prevent precommit gotchas

### LINKED ISSUE #429: autopep8 is so annoying
## Describe the bug
[The maximum line length autopep8 allows is 79 by default](https://github.com/hhatto/autopep8#usage), and it's so annoying. Since we don't enforce the line limit on GitHub Actions, code on which autopep8 was not run *can* pass the CI. The problem arises when someone else runs `precommit` only to find that a previous developer has not run `precommit` (which includes `autopep8`), in which case we have unintended formatting changes.

## To Reproduce
Steps to reproduce the behavior. For example:
1. Run `docker exec tcf_django autopep8 --in-place --jobs=0 --aggressive --aggressive -r tcf_website tcf_core` (which is the same as what's in `precommit` at this point).
2. Run `git status`, and verify that there are unintended formatting changes.

## Expected behavior
`autopep8` shouldn't wrap around if the line isn't too long.

## Screenshots
image

## Additional context
People may have different opinions on the best line limit, but 80 is too small.

Bartleby2718:
## Status
- Done

- This PR closes #429 and prevents a common mistake when editing shell scripts.

## What I did
- Changed `precommit` to a shell script and added `.gitattributes`

## Tests
- Run `echo 1\r\n > echo.sh` and then `git add .` You should see `warning: CRLF will be replaced by LF in echo.sh.` because of `\r\n` in the shell script. In some cases, your shell script won't work as intended, but only in your working directory. It own't be committed to the repository.

- Source: https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings
- A caveat is that we'll have one more file in the root. Is this okay?

## Merging the PR
- Who should merge this PR?
  - [ ] I will merge it myself
    - [ ] preserve the history
    - [x] squash-merge


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Declare files that will always have LF line endings on checkout.
```
alex-shen1, line 1: Maybe leave a comment going more into detail why this is necessary, for when someone sees this several years from now and wonders why we have a random 2-line file.

precommit.sh
```diff
@@ -1,5 +1,6 @@
+#!/bin/sh
 # Auto format.
-docker exec tcf_django autopep8 --in-place --jobs=0 --aggressive --aggressive -r tcf_website tcf_core
+docker exec tcf_django autopep8 --in-place --jobs=0 --max-line-length=100 --aggressive --aggressive -r tcf_website tcf_core
```
alex-shen1, line 3: How did you decide exactly 100 characters was the right amount? (side note, 79 is such a weird default limit)

alex-shen1 (changes requested):

.gitattributes
Bartleby2718: Fixed in a57c480866f30b32cb3248a25bec775e9d33b02f

precommit.sh
Bartleby2718: I believe 80, 100, and 120 are some popular choices--[this is the only source I could find, though](https://www.quora.com/Why-does-Google-Java-coding-style-guide-chose-maximum-of-100-characters-column-limit-per-line-why-not-120-just-curious-to-know)--but based on our reaction to `autopep8`'s autoformatting, I think there is a consensus that 80 is too short. Also, 120 is not too long for widescreen monitors, but given that most of us are students developing on a laptop, I chose 100. Does this make sense?

As a side note, [79 is mentioned in Python PEP.](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

alex-shen1: Sounds good to me!

alex-shen1 (approved):

PR #434, opened 2021-03-13
Grades Data overhaul

### LINKED ISSUE #16: Grades data

### LINKED ISSUE #233: Overhaul + improve grades data
Merging in today for launch, but it's kinda wack so we should revisit it at some points. There are gonna be comments from #209 that are unaddressed that we should go back to look at.

alex-shen1:
## Status
- Done (maybe)

## GitHub Issues addressed
- This PR closes #16, closes #233 

## What I did
- Got FOIA grades data from UVA + added it to the repo, split by semester, then updated our `load_grades` command to match the new format
- Deleted the following fields from CourseGrade and CourseInstructorGrade (they're redundant and can be accessed from the foreign keys):
image
- Deleted some fields from CourseGrade and CourseInstructorGrade because our new data doesn't have them (kinda sucks but whatevs)
image
- Discovered (and fixed) a MAJOR BUG where data lists were being copied by reference instead of by value, so when `CourseInstructorGrade` got incremented, so did `CourseGrade`. This means that **every piece of data on our site right now is probably wrong.** This shows the importance of writing tests, people!
- Minor refactors/documentation
- Had to add `pandas` and `tqdm` to the installed packages list on GitHub Actions to get it to work. This will increase build times by like 2 seconds. Sorry not sorry.

## Tests
- Large-scale test: Pull this branch and run `docker exec tcf_django python manage.py load_grades ALL_DANGEROUS`. Might break your data.
- Alternatively, I've created a set of automated tests in `test_commands.py` 

- I think I figured out part of why we had a bunch of missing data - beyond checking first/last name, the old script also checked email, and there may sometimes have been discrepancies between our DB and the grade data as to whether or not they both had/both didn't have an email for a particular professor.
- The new data lumps everything below a C- into one category `DFW`, which I assumed to mean "drop, fail, or withdraw." The name I gave this field in the code/our updated models is `not_pass`, which is admittedly pretty bad. If anyone has better ideas, I'd be more than willing to adopt them.
- This might also affect our average GPA calculation since we can't differentiate fails and withdraws (which I don't think are included on GPA calculations? This is a draft, so I'm writing this as a reminder for myself to check this manually against the data later)
- Looks like the command has a default verbosity of 1? I thought commands had it to 0 by default...is that intended?
- I've decided to commit the migrations this time because it's fairly large and I imagine as things get more complicated, we should start adhering to the general best practice of doing so. LMK if you disagree.
- Looks like Django commands aren't being checked by Pylint. Is this OK, or should we change that?
- It might be unnecessary to have the original data and the `split_data.py` script I used to split it into the individual semester files. Should I just delete this?
- The script has a *critical* assumption, being that no two instructors share the exact same first/last name. We _could_ add middle name (but the grades data has a mix of middle initial and full middle name) to reduce the probability of an overlap, but that doesn't seem very robust either.... unfortunately the grades data UVA provides doesn't include computing ID.

## Merging the PR
- Who should merge this PR?
  - [x] I will merge it myself
  - [ ] The last reviewer to approve can merge it
  - [x] preserve the history
  - [ ] squash-merge
 
I've been squashing overlapping commits locally, so each of the commits here covers a separate "thing".


Bartleby2718 (changes requested):
Stylistic review! Feel free to dismiss the ones you don't think are necessary.

tcf_website/management/commands/grade_data/split_data.py
```diff
@@ -0,0 +1,17 @@
+"""
+The data came from UVA was 2009-2020, but that's bulky and we want it to come
+in semester-by-semester files. This script probably won't ever be run again, but
+may come in handy if we ever need to break up a big data file by semester.
+"""
+from pandas import read_csv
+
+raw_data = read_csv("csv/GradeDistributionF09toSpr20.csv", encoding="ISO-8859-1")
```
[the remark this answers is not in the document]
alex-shen1: Yeah, the default (UTF-8?) broke so I just randomly Googled and threw encodings into it until it worked lol

tcf_website/static/course/javascript/course_data.js
```diff
@@ -96,7 +88,11 @@ const createChart = gradesData => {
                     label: function(tooltipItem, data) {
                         var dataset = data.datasets[0];
                         var percent = Math.round((dataset.data[tooltipItem.index] / dataset._meta[0].total) * 100);
-                        return data.labels[tooltipItem.index] + ": " + percent + "%";
+                        var label = data.labels[tooltipItem.index];
+                        if (tooltipItem.index === 9) {
+                            label = "No Credit";
```
Bartleby2718, line 93: Will `No Credit` be enough for users? I agree it's the most accurate term, but I'm afraid it might not be the most intuitive one.

tcf_website/tests/test_commands.py
```diff
@@ -0,0 +1,69 @@
...
+        super().setUpClass()
+        setup(cls)
+        management.call_command('load_grades', 'test_data', '--suppress-tqdm', verbosity=0)
+        cls.cg = CourseGrade.objects.all()[0]
+        cls.cig = CourseInstructorGrade.objects.all()[0]
+
+    def test_no_duplicates(self):
+        """Make sure only one instance of CourseGrade and CourseInstructorGrade were created"""
+        self.assertEqual(len(CourseGrade.objects.all()), 1)
+        self.assertEqual(len(CourseInstructorGrade.objects.all()), 1)
```
Bartleby2718, line 24:
```suggestion
        self.assertEqual(CourseGrade.objects.count(), 1)
        self.assertEqual(CourseInstructorGrade.objects.count(), 1)
```

tcf_website/tests/test_commands.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(self.cg.course, self.cig.course)
+
+    def test_correct_instructor(self):
+        """Make sure instructor is Tom Jefferson"""
+        self.assertEqual(self.cig.instructor.full_name(), "Tom Jefferson")
+
+    def test_total_students(self):
+        """Check valid total_enrolled"""
+        self.assertEqual(self.cg.total_enrolled, self.cig.total_enrolled)
+        self.assertEqual(self.cg.total_enrolled, 24)
```
Bartleby2718, line 37:
```suggestion
        self.assertEqual(self.cg.total_enrolled, 24)
        self.assertEqual(self.cig.total_enrolled, 24)
```
I think this would be easier to debug.

tcf_website/tests/test_commands.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(self.cig.a_plus, self.cig.a_plus)
+        self.assertEqual(self.cg.a, self.cig.a)
+        self.assertEqual(self.cg.a_minus, self.cig.a_minus)
+        self.assertEqual(self.cg.b_plus, self.cig.b_plus)
+        self.assertEqual(self.cg.b, self.cig.b)
+        self.assertEqual(self.cg.b_minus, self.cig.b_minus)
+        self.assertEqual(self.cg.c_plus, self.cig.c_plus)
+        self.assertEqual(self.cg.c, self.cig.c)
+        self.assertEqual(self.cg.c_minus, self.cig.c_minus)
+        self.assertEqual(self.cg.no_credit, self.cig.no_credit)
```
Bartleby2718, line 69: I wonder if it's worth replacing this with a for loop:
```
for field in ['a_plus','a','a_minus','b_plus','b','b_minus','c_plus','c','c_minus','no_credit']:
    cg_field = getattr(self.cg, field)
    cig_field =getattr(self.cig, field) 
    self.assertEqual(cg_field , cig_field)
```

tcf_website/management/commands/load_grades.py
```diff
@@ -67,48 +105,63 @@ def clean(self, df):
         CourseGrade.objects.all().delete()
         CourseInstructorGrade.objects.all().delete()
```
Bartleby2718, line 106: Because `clean` truncates existing these two tables and `load_semester_file` calls `clean`, the entire script works only for `ALL_DANGEROUS`. For example, if you run `docker exec -ti tcf_django python3 manage.py load_grades 2018_spring` and then run `docker exec -ti tcf_django python3 manage.py load_grades 2018_fall`, Spring 2018 data will be wiped out before loading Fall 2018. 

FYI This was a bug from my code, but I was unable to fix this. I know it's a little harsh to have you fix this. As a last resort, we can simply move these two lines and just not support adding a single semester, since redoing everything from scratch isn't so slow.

tcf_website/management/commands/grade_data/csv/test_data.csv
```diff
@@ -0,0 +1,3 @@
+Term Desc,Subject,Catalog Number,Class Title,Primary Instructor Name,Class Section,Course GPA,# of Students,A+,A,A-,B+,B,B-,C+,C,C-,DFW
+2009 Fall,CS,420,Software Testing,"Jefferson,Tom",X,X,15,0,2,5,3,0,0,1,1,0,3
+2009 Fall,CS,420,Software Testing,"Jefferson,Tom",X,X,9,1,2,1,0,0,3,0,1,0,1
```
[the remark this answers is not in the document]
Bartleby2718: In particular, this causes the load_grades script to load all data basically twice.

Bartleby2718 (changes requested):
I just spent 1.5 hours on this review, but I still haven't got to `load_grades`. Since I've identified a serious bug, I'll submit a PR now and take a look later.

tcf_website/management/commands/grade_data/csv/test_data.csv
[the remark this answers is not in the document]
alex-shen1: Oh yeah, I realized this was the case but didn't get around to removing it. I'll put it in the Google Drive when I do.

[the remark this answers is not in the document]
alex-shen1: Also, I'm realizing that `test_data` definitely needs to be moved out because as it is, the dummy data will end up being loaded into prod yikes

tcf_website/management/commands/load_grades.py
alex-shen1: Yeah, I found that to be a little bit weird when I was working, but wasn't sure if that was within the scope of the PR because it was already there before and I figured we reload everything all the time anyways.

Should probably open a new issue for it at the very least; am I correct in assuming that the solution is simply a matter of filtering all `CourseGrade`/`CourseInstructorGrade` instances by semester and deleting?

tcf_website/static/course/javascript/course_data.js
```diff
@@ -1,10 +1,9 @@
 const loadData = data => {
     // order in the input data
     /* eslint-disable camelcase */
-    const { a_plus, a, a_minus, b_plus, b, b_minus, c_plus, c, c_minus, d_plus, d, d_minus, f, ot, drop, withdraw } = data;
+    const { a_plus, a, a_minus, b_plus, b, b_minus, c_plus, c, c_minus, no_credit } = data;
```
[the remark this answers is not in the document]
alex-shen1: I thought it just used the data from the `CourseInstructorGrade` it got from the view? We create a variable `grades_data` and pass its fields in the context variable `data`. In any case, I should fix the API.

tcf_website/static/course/javascript/course_data.js
alex-shen1: I thought the same, but I'm not sure where to add an elaboration on what it means because you'd have to have like a tooltip on a tooltip or have it somewhere not on the chart. Any ideas?

Bartleby2718: I thought it would be okay to add more in the parentheses following that, but we should probably ask the team.

tcf_website/management/commands/load_grades.py
Bartleby2718: Yeah, I think so. You just need to make sure it's idempotent (i.e. loading the same semester multiple times should have the same effect as loading once successfully).

tcf_website/migrations/0003_auto_20210330_2024.py
```diff
@@ -0,0 +1,23 @@
...
+            model_name='coursegrade',
+            old_name='not_pass',
+            new_name='no_credit',
+        ),
+        migrations.RenameField(
+            model_name='courseinstructorgrade',
+            old_name='not_pass',
+            new_name='no_credit',
+        ),
+    ]
```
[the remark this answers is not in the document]
alex-shen1: Will do! - it ended up being like this because the field got renamed after I had made the original migration files, and I never got around to changing it.

alex-shen1:
TODO for me:
- Remove the big data CSV file + put it in Google Drive
- Add an exception in the command to make sure the test data CSV doesn't get added to prod DB
- Delete migration 003 and 002; remake migrations to consolidate changes in one file
- Add new line to bottom of `.gitignore`
- Add explanation of `split_data.py` to Wiki
- Discuss the NC label @ meeting
- Commit @Bartleby2718's suggestions

tcf_website/tests/test_commands.py
alex-shen1: Technically the CSV data file is CS 420, but I'll just update the CSV to reflect this because we want 4-digit class codes.

tcf_website/tests/test_commands.py
```diff
@@ -0,0 +1,69 @@
...
+        cls.cg = CourseGrade.objects.all()[0]
+        cls.cig = CourseInstructorGrade.objects.all()[0]
+
+    def test_no_duplicates(self):
+        """Make sure only one instance of CourseGrade and CourseInstructorGrade were created"""
+        self.assertEqual(len(CourseGrade.objects.all()), 1)
+        self.assertEqual(len(CourseInstructorGrade.objects.all()), 1)
+
+    def test_correct_course(self):
+        """Make sure the course for both is CS 420"""
```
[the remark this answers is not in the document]
alex-shen1: Technically the CSV data file is CS 420, but I'll just update the CSV to reflect this because we want 4-digit class codes.

tcf_website/tests/test_commands.py
alex-shen1: I'll do it!

Bartleby2718 (commented):
More comments on the topics you brought up:
1. Including `email` as part of the key
This was my deliberate (and conservative) choice to be future-proof, but I acknowledge that it leaves out some of the data. I see you've written in the comment that now UVA no longer provides us with the email of the primary instructor. When we submit a FOIA request next time, do you think it's better to ask them to include the emails? What's also concerning for me is that they might change the format again next time, which makes it harder to reuse the script.

2. The default verbosity is 1 [according to the documentation](https://docs.djangoproject.com/en/3.1/ref/django-admin/#cmdoption-verbosity).

3. I'm seeing generally higher GPAs for most classes, and it seems like this is because we excluded DFW from our GPA calculations. Should we include an alert that says something like "We've updated how average GPAs are calculated, due to the change in the underlying data provided by UVA."?

4. Can we have a patch that includes only this PR? Since we're including migration files for the first time (a year, probably), I want to make sure this doesn't break anything.

5. I don't see any reason why we shouldn't run `pylint` on the Django commands.

tcf_website/management/commands/grade_data/csv/test_data.csv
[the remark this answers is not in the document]
alex-shen1: I decided to just hardcode a check for `file != 'test_data.csv'` in the script to make the command test easier. Is this okay?

alex-shen1:
1. Darn, totally forgot about the FOIAs. I'll submit a new one ASAP. Not sure if they'll actually follow through, but it can't hurt to file a new request.
2. 👍 (this is here because markdown won't let me skip the list to 3 lol)
3. I'm personally skeptical that people will care that much, but I suppose it couldn't hurt to add it. How long do you think it should stick around before we take it back down? (Also, it might interfere with the ads Jules is working on?)
4. Yeah, sure.
5. I agree; I'm not exactly sure why it's not being run already by `precommit`. Not looking forward to fixing all this wack code though, pylint scores it really poorly :/

.gitignore
```diff
@@ -28,3 +25,5 @@ db/*.csv
 .coverage
 # ESLint cache
 .eslintcache
+# Logging text files
+*.txt
```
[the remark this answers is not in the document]
alex-shen1: Done

alex-shen1:
BUG: course-instructor crashes when there is no grade data available

image

@ http://localhost:8000/course/764/8083

alex-shen1:
@Bartleby2718 do you know if we currently use the API on the site specifically for the grade categories? To my knowledge, the API helps with pagination but that only shows average stats. Basically I'm asking if the API fix has to be in this PR lol

Also, (besides the obvious school commitments) I waited to work on this until UVA FOIA got back to me with new data, but....the new data was the same as the old data. Looks like we can't use emails after all. The new data also oddly contains some additional columns, but the ones we actually care about are unmodified. I'll write up more comprehensive information on the wiki.

alex-shen1:
@Bartleby2718 do you know if we currently use the API on the site specifically for the grade categories? To my knowledge, the API helps with pagination but that only shows average stats. Basically I'm asking if the API fix has to be in this PR lol

Also, (besides the obvious school commitments) I waited to work on this until UVA FOIA got back to me with new data, but....the new data was the same as the old data. Looks like we can't use emails after all. The new data also oddly contains some additional columns, but the ones we actually care about are unmodified. I'll write up more comprehensive information on the wiki.

.gitignore
Bartleby2718, line 29: Technically this will also ignore `requirements.txt` and `runtime.txt`, although they won't be ignored because they were created before the change was made. I think a better practice is to specifically prevent ignoring those files (with leading exclamation marks) or ignore specific log files (e.g. `log/*.txt`).

tcf_website/tests/test_commands.py
```diff
@@ -57,13 +57,17 @@ def assert_correct_data(self, model):
...
+            'a_plus',
+            'a',
+            'a_minus',
+            'b_plus',
+            'b',
+            'b_minus',
+            'c_plus',
+            'c',
+            'c_minus',
+                'no_credit']:
```
Bartleby2718, line 70: Indent is off for some reason here.

Bartleby2718 (changes requested):

Bartleby2718:
> 3. I'm personally skeptical that people will care that much, but I suppose it couldn't hurt to add it. How long do you think it should stick around before we take it back down? (Also, from a UI perspective, having another banner at the top might interfere with the ads Jules is working on?)

People will have forgotten everything by the summer break is over, so I guess we can do away with it.

> @Bartleby2718 do you know if we currently use the API on the site specifically for the grade categories? To my knowledge, the API helps with pagination but that only shows average stats. Basically I'm asking if the API fix has to be in this PR lol

Not sure what you mean by "grade categories," but I just pushed f843acc1174add0502b89073243aed4718670a9e to makes necessary changes to the Course API.

tcf_website/api/serializers.py
```diff
@@ -86,13 +86,7 @@ class CourseAllStatsSerializer(CourseSimpleStatsSerializer):
...
     c = serializers.IntegerField(allow_null=True)
     c_minus = serializers.IntegerField(allow_null=True)
-    d_plus = serializers.IntegerField(allow_null=True)
-    d = serializers.IntegerField(allow_null=True)
-    d_minus = serializers.IntegerField(allow_null=True)
-    f = serializers.IntegerField(allow_null=True)
-    ot = serializers.IntegerField(allow_null=True)
-    drop = serializers.IntegerField(allow_null=True)
-    withdraw = serializers.IntegerField(allow_null=True)
+    no_credit = serializers.IntegerField(allow_null=True)
```
joek13, line 89: I think `no_credit` is a fine internal name as long as the UI tooltip shows something like "Drop/Fail/Withdraw" to be more descriptive

tcf_website/management/commands/load_grades.py
```diff
@@ -118,159 +179,135 @@ def load_row_into_dict(self, row):
...
+            'b_minus': data[5],
+            'c_plus': data[6],
+            'c': data[7],
+            'c_minus': data[8],
+            'no_credit': data[9],
+            'total_enrolled': total_enrolled
+        }
+
+        if has_instructor:
+            course_grade_params['instructor_id'] = self.instructors.get(row[2:])
```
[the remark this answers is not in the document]
joek13: actually, it looks like the [way we create `Instructor` objects](https://github.com/thecourseforum/theCourseForum2/blob/267797b032918357e8abc3409eb3de70c7009724/tcf_website/management/commands/load_semester.py#L234-L237) fundamentally allows only one instructor with a given first/last name, so this only makes a difference in the rare case of when we have manually split a professor into two. (i believe that is something we can do!)

in fact, I ran this query to check how unique `Instructor` objects we have for each first/last name:
```sql
SELECT
    COUNT(DISTINCT tcf_website_instructor.id) as count,
    tcf_website_instructor.last_name,
    tcf_website_instructor.first_name
FROM
    tcf_website_instructor
GROUP BY
    tcf_website_instructor.last_name,
    tcf_website_instructor.first_name
ORDER BY count DESC;
```

...and it appears that there is a one-to-one correspondence between (first name, last name) pairs and `Instructors`. so it looks like this change would have no effect presently, even if it is in principle more robust.

joek13 (dismissed):
my main nitpick is with "NC" being used as a label instead of something more informative, but otherwise, lgtm!

tcf_website/management/commands/load_grades.py
```diff
@@ -118,159 +179,135 @@ def load_row_into_dict(self, row):
...
-                    len(course_instructor_grades[course_instructor_identifier])):
-                course_instructor_grades[course_instructor_identifier][i] += this_semesters_grades[i]
+        # Load this semester into course_instructor_grades dictionary
+        if course_instructor_identifier in self.course_instructor_grades:
+            for i in range(len(self.course_instructor_grades[course_instructor_identifier])):
+                self.course_instructor_grades[course_instructor_identifier][i] += this_semesters_grades[i]
         else:
-            course_instructor_grades[course_instructor_identifier] = this_semesters_grades
+            self.course_instructor_grades[course_instructor_identifier] = this_semesters_grades.copy(
+            )
```
jackliu612, line 215: minor styling comment about parens on different lines

tcf_website/management/commands/load_grades.py
```diff
@@ -118,159 +179,135 @@ def load_row_into_dict(self, row):
...
             print('Step 2: Bulk-create CourseGrade instances')
-        # used for gpa calculation
+        # Used for GPA calculation
         grade_weights = [4.0, 4.0, 3.7,
                          3.3, 3.0, 2.7,
                          2.3, 2.0, 1.7,
-                         1.3, 1.0, 0.7,
-                         0,
-                         0, 0, 0]
+                         0]
```
jackliu612, line 230: Are grade calculations are changed by a large amount based on counting dfw as 0? Slightly worried if a class has a sizable amount of withdraws this will impact the average significantly.

jackliu612: nvm I see that later on in the helper function you remove the DFW from the denominator. Maybe make a comment about it? also fine as is.

tcf_website/management/commands/load_grades.py
```diff
@@ -118,159 +179,135 @@ def load_row_into_dict(self, row):
...
         if self.verbosity > 0:
             print('Done creating CourseInstructorGrade instances')
+
+        if self.log_missing_instructors:
+            print('Writing missing instructors to missing-instructors.txt')
+            with open('missing-instructors.txt', 'w') as file:
+                for instructor in self.missing_instructors:
+                    file.write(instructor)
+
+    def set_grade_params(self, row, total_enrolled, total_weight, has_instructor):
```
jackliu612, line 277: `has_instructor` is a bit of a misleading parameter that can be confused with the missing instructors list. Instead maybe `is_instructor_grade` or something like that?

jackliu612 (commented):
Initial review. Only looked at the loading script, and did not test locally yet.

tcf_website/api/serializers.py
```diff
@@ -85,13 +85,7 @@ class CourseAllStatsSerializer(CourseSimpleStatsSerializer):
...
     c = serializers.IntegerField(allow_null=True)
     c_minus = serializers.IntegerField(allow_null=True)
-    d_plus = serializers.IntegerField(allow_null=True)
-    d = serializers.IntegerField(allow_null=True)
-    d_minus = serializers.IntegerField(allow_null=True)
-    f = serializers.IntegerField(allow_null=True)
-    ot = serializers.IntegerField(allow_null=True)
-    drop = serializers.IntegerField(allow_null=True)
-    withdraw = serializers.IntegerField(allow_null=True)
+    dfw = serializers.IntegerField(allow_null=True)
```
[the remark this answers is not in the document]
alex-shen1: Now that you mention it, I think you're right. I sent an email to the FOIA people at UVA to ask for clarification, but until they get back to me with a different answer I'll change the labels to `D/F/Withdraw`.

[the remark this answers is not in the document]
alex-shen1: They said it's D, F, withdraw. I'm gonna update the label accordingly.

tcf_website/management/commands/load_grades.py
```diff
@@ -15,16 +14,50 @@ class Command(BaseCommand):
...
-    course_instructor_grades = {}
+        # Initialize some variables that get used everywhere
+
+        # Dict mapping instructor first/last name to their ID in our database
+        # We used to be able to include email, but the UVA dataset doesn't include it!
+        # This is pretty risky if instructors have the exact same name, so yikes
+        # TODO: figure out how to differentiate (maybe by classes taught?)
+        self.instructors = {
+            (obj['first_name'], obj['last_name']): obj['id']
+            for obj in Instructor.objects.values('id', 'first_name',
```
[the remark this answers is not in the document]
alex-shen1: Looks like [Joe looked into this](https://github.com/thecourseforum/theCourseForum2/pull/434#discussion_r779214887) and it seems like the way our load semester script works makes it impossible for there to be duplicates (because instructors with the same name are all consolidated into the same Instructor object lol). The SIS API does support providing instructor emails, so when (/if lol) that gets reworked, then we'll have to revisit this.

tcf_website/management/commands/load_grades.py
alex-shen1: Resolved

tcf_website/management/commands/load_grades.py
```diff
@@ -37,78 +70,106 @@ def add_arguments(self, parser):
...
+            CourseGrade.objects.all().delete()
+            CourseInstructorGrade.objects.all().delete()
+
+            # Loads every data CSV file in /grade_data/csv with exceptions
             for file in sorted(os.listdir(self.data_dir)):
-                if file[0] not in ('.', '~'):
+                if self.verbosity == 3:
+                    print('Loading data from', file)
+                # Ignore temp files (start with '~' on Windows, '.' otherwise) and test data
+                if file[0] not in ('.', '~') and file != 'test_data.csv':
```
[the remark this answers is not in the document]
alex-shen1: In either case, I think you'd have to hardcode _something_ related to the test file's path, since `manage.py load_grades {FILENAME}` expects `FILENAME` to be located in the `csv` folder, so if we made a separate directory the test command would be something like `manage.py load_grades ../test/test_data.csv` which also doesn't feel great. Arguably that refactors the jank into test code that isn't looked at as often...but I don't think that's a great approach either.

jackliu612 (approved):
Thanks for addressing my review. I trust that you will not break our site :)

PR #417, opened 2021-02-28
Changed review model and view to display both upvotes and downvotes

andrewmniu:
## Status
- Done

## GitHub Issues addressed (`#IssueNumber`)
- This PR closes n/a

## What I did
- Made minor changes to change the voting system to a like system

## Screenshots
- Before 
image

- After
image

## Tests
- I've yet to write tests

## Questions/Discussions/Notes
- I may need a design mockup for the updated review card if we choose to change to this liking system
- I was struggling to modify the JavaScript in `review.js` to actually handle this voting system, so I might need help on that
- Sorting by helpfulness now sorts by most upvotes, so I also might need help if we still want to sort by net upvotes/downvotes.

## Merging the PR
- Who should merge this PR?
  - [ ] I will merge it myself
  - [x] The last reviewer to approve can merge it
- Should the commit history be preserved in the base branch, or is it okay to combine all of them into a single commit ("squash-merge")?
  - [ ] preserve the history
  - [x] squash-merge

PR #443, opened 2021-03-14
Emoji reorder

### LINKED ISSUE #381: Review Page Emojis
**Is your feature request related to a problem? Please describe.**
The emojis are at the top of the page are in a different order than those in the reviews. There is also ambiguity in the images.

Reorder the emojis so they're consistent. It would also be nice to have some way to reiterate the meaning of the emotions in each review. 

**Describe alternatives you've considered**
N/A 

**Additional context**

quyenxhuynh:
## Status
- Done? 

## GitHub Issues addressed
- This PR closes #381 

## What I did
- Changed order of emojis
- Added tooltips (and changed CSS for tooltips-inner)
- Changed orders of recommend and difficulty because the average rating does not consider the difficulty rating, but only the other three variables

## Screenshots
- Before

Screen Shot 2021-03-14 at 1 36 21 PM

- After
- 
Screen Shot 2021-04-11 at 2 32 11 PM

## Tests
Go to any course page and check that emojis orders match on the instructor card and each individual review
Check tooltips (hover over review emojis or hold down on mobile) to see if they appear

- Tooltips on mobile is a little unintuitive, but I assume most people aren't really looking at tCF on their phone so not a super big issue?
- Emojis still kind of ambiguous on their own - wasn't sure how to improve it much further though

## Merging the PR
- Who should merge this PR?
  - [ ] I will merge it myself
  - [ ] preserve the history
  - [x] squash-merge


n4mahajan:
Nice job! I really like the tooltip, I think that's a nice addition. My main suggestion I have is perhaps to switch the order of "Difficulty" and "Recommendability"? This would apply to both a review card for a given course_instructor page and the order  when someone writes a new review so that the order is consistent like you made it now. The reason I suggest to change the order is because the average review score is calculated based on the instructor rating, enjoyability, and recommendability and the difficulty is not part of the calculation. Thus, I think having difficulty third in the order and in between the other three makes it less intuitive/detracts from how the average score is actually calculated, so I would suggest making the difficulty last for all of the relevant orders. 

Also, to stay consistent with what it actually says when you're writing a review, maybe the tooltip for "Instructor Rating" should be changed to "Instructor", "Recommendability" to "Recommend", and "Review Average" to "Review Score" (this mainly because it's not immediately clear what the review is an average of)?

alex-shen1 (changes requested):
To echo what's already been said, good job! The tooltips are definitely a nice addition. To chime in on the conversation about specifically what should be on those tooltips, I think that the labels you came up with are honestly better than what's on the review form currently, so we might want to think about renaming that, maybe in another PR later? (@n4mahajan good catch noticing the inconsistency!) Also, I think "Average Rating" makes sense for the average stat, considering that we already have the format of "X Rating" with instructor and that the number itself is just an arithmetic mean of the three values.

quyenxhuynh:
Update 3/25: Switched Difficulty and Recommendability per @n4mahajan's request and also changed the tooltip labels to match the card at the top of the review page. Wasn't able to change the card (since Alex mentioned my labels were more informative) due to the bar chart, so maybe we can change that in a different PR. Probably also want a new PR to change it everywhere else these emojis may appear (new review? maybe that's it?)

j-alicia-long:
The updates look good! 

@andrewmniu is doing an overhaul of the review card layout in #417 including swapping out the icon for "Recommend", so we should probably wait for that to be merged in.

j-alicia-long:
@quyenxhuynh once you resolve the merge conflicts I can approve this again!

j-alicia-long (approved):

PR #452, opened 2021-03-21
Display total number of reviews for a given course (with or without t…

n4mahajan:
…ext)

## Status

- Done?

## GitHub Issues addressed

- This PR closes

## What I did

- Implemented Lilleth's mockup to display the total number of reviews (with or without text) for a given course on its page.

## Screenshots

- Before
Screen Shot 2021-03-21 at 1 55 04 AM
Screen Shot 2021-03-21 at 2 18 46 AM

- After
Screen Shot 2021-03-21 at 3 14 19 PM

Screen Shot 2021-03-21 at 3 14 48 PM

## Tests

- Go to various courses to make sure that the total number of reviews shows. Ideally, we can verify that the total number of reviews matches up with how many reviews the course actually has but I'm not sure how to verify this as of yet.
- Also, verify that courses with only 1 review and/or comment say "1 Review" and "1 Comment" as opposed to "Reviews" and "Comments".

- Not sure if we want to change the display/message for courses with 0 reviews or if we want to do anything special there?
- Also, getting some fails with pylint because of "too many local variables" in browse views?

## Merging the PR

- Who should merge this PR?
  - [] I will merge it myself
  - [] preserve the history
  - [x] squash-merge

### DIFF
--- a/tcf_website/templates/course/course_professor.html
+++ b/tcf_website/templates/course/course_professor.html
@@ -61,7 +61,7 @@
                 <div class="ratings-row">
                     <div class="rating-card">
                         <span class="rating-card-num rating-num">&mdash;</span>
-                        <p class="rating-card-desc"><i class="fa fa-star fa-fw"></i>&nbsp;Rating</p>
+                        <p class="rating-card-desc"><i class="fa fa-star fa-fw"></i>&nbsp;{{ num_reviews }} Reviews</p>
                     </div>
                     <div class="info-container">
                         <div class="info-row">
--- a/tcf_website/templates/course/course_professor.html
+++ b/tcf_website/templates/course/course_professor.html
@@ -61,7 +61,7 @@
                 <div class="ratings-row">
                     <div class="rating-card">
                         <span class="rating-card-num rating-num">&mdash;</span>
-                        <p class="rating-card-desc"><i class="fa fa-star fa-fw"></i>&nbsp;{{ num_reviews }} {% if num_reviews == 1 %} Review {% else %} Reviews {% endif %}</p>
+                        <p class="rating-card-desc"><i class="fa fa-star fa-fw"></i>&nbsp;{{ num_reviews }} {% if num_reviews == 1 %} Rating {% else %} Ratings {% endif %}</p>
                     </div>
                     <div class="info-container">
                         <div class="info-row">
--- a/tcf_website/templates/course/course_professor.html
+++ b/tcf_website/templates/course/course_professor.html
@@ -193,7 +193,7 @@
 
         <div class="reviews-container mb-5 px-3">
             <div class="reviews-toolbar pb-2 d-flex flex-md-row flex-column justify-content-left justify-content-md-center">
-                <h3 class="reviews-heading mb-0 mt-1 mr-4">{{ reviews|length }} Comments</h3>
+                <h3 class="reviews-heading mb-0 mt-1 mr-4">{{ reviews|length }} {% if reviews|length == 1 %} Comment {% else %} Comments {% endif %}</h3>
                 <div class="reviews-options d-flex flex-sm-row flex-column col p-0">
                     <a href="{% url 'new_review' %}" class="btn bg-tcf-orange text-white mb-3">
                       <i class="fa fa-plus" aria-hidden="true"></i> Add your review!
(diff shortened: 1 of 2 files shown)


borjainglefernandez:
I took a look and it seems to be working great! This is definitely a good feature to implement so that people know how many reviews there are. I have doubts about the position of the number of reviews. I personally do not think we should get rid of the "Rating" under the rating since it can be confusing.

tcf_website/templates/course/course_professor.html
```diff
@@ -61,7 +61,7 @@ <h2 class="mr-md-3">{{ course.title }}</h2>
                 <div class="ratings-row">
                     <div class="rating-card">
                         <span class="rating-card-num rating-num">&mdash;</span>
-                        <p class="rating-card-desc"><i class="fa fa-star fa-fw"></i>&nbsp;Rating</p>
+                        <p class="rating-card-desc"><i class="fa fa-star fa-fw"></i>&nbsp;{{ num_reviews }} {% if num_reviews == 1 %} Review {% else %} Reviews {% endif %}</p>
```
borjainglefernandez, line 64: Since we already add the number of comments on line 196, wouldn't it be better to just keep this as Rating?

tcf_website/templates/course/course_professor.html
```diff
@@ -193,7 +193,7 @@ <h2 class="mr-md-3">{{ course.title }}</h2>
 
         <div class="reviews-container mb-5 px-3">
             <div class="reviews-toolbar pb-2 d-flex flex-md-row flex-column justify-content-left justify-content-md-center">
-                <h3 class="reviews-heading mb-0 mt-1 mr-4">{{ reviews|length }} Comments</h3>
+                <h3 class="reviews-heading mb-0 mt-1 mr-4">{{ reviews|length }} {% if reviews|length == 1 %} Comment {% else %} Comments {% endif %}</h3>
```
borjainglefernandez, line 196: Just in general since we're eventually moving to be able to comment on reviews, it might be best to change Comments to Reviews?

borjainglefernandez: Oh wait I see, this displays total comments (reviews and comments) whereas the above displays number of reviews right?

tcf_website/templates/course/course_professor.html
borjainglefernandez: I still think this should be kept as rating and we can move number of reviews somewhere else

tcf_website/templates/course/course_professor.html
```diff
@@ -193,7 +193,7 @@ <h2 class="mr-md-3">{{ course.title }}</h2>
 
         <div class="reviews-container mb-5 px-3">
             <div class="reviews-toolbar pb-2 d-flex flex-md-row flex-column justify-content-left justify-content-md-center">
-                <h3 class="reviews-heading mb-0 mt-1 mr-4">{{ reviews|length }} Comments</h3>
```
borjainglefernandez, line 196: What's the difference between reviews|length and num_reviews?

borjainglefernandez:
Screenshot 2021-03-21 at 11 00 57 AM
Just a thought, but I think it might be better to include the number of total reviews (written and not written) in the middle of this header as opposed to getting rid of rating altogether.

n4mahajan:
> Screenshot 2021-03-21 at 11 00 57 AM
> Just a thought, but I think it might be better to include the number of total reviews (written and not written) in the middle of this header as opposed to getting rid of rating altogether.

Something like this? 
Screen Shot 2021-03-21 at 3 39 04 PM

I didn't really style it or anything and it would probably need its own section/box to fit in better, but your idea could work?

alex-shen1:
I think it definitely shouldn't be centered like that, but it could work on one of the sides?

n4mahajan:
> I think it definitely shouldn't be centered like that, but it could work on one of the sides?

Will try putting it under the prof name as Lilleth suggested, but otherwise I think it's best leaving it like this unless we really want to preserve the "Rating" label

alex-shen1:
> also think since we are eventually implementing a commenting feature, we should change "Comment(s)"to "Written Review(s)".

Also to address this particular concern, my understanding was that we would include these subcomments/replies/whatever we end up calling them in this number?

alex-shen1 (approved):
I think this is the best it's gonna get for now - I'm still a bit iffy on expecting users to be able to infer things, but at the very least when the emoji tooltips get merged that might help, and it certainly looks good. Nice job catching the n=1 edge case too lol

PR #459, opened 2021-03-21
Current semester redundancy

### LINKED ISSUE #455: Get rid of last taught semester redundancy
“Last taught Spring 2021” is redundant if filtering by all classes for Spring 2021, get rid of the redundancy

khushic:
## Status

- Done

## GitHub Issues addressed

- This PR closes #455 

## What I did

- Added a css class that changes the opacity of the last taught section to 0 when the latest semester/all button is toggled in show-hide.js

## Screenshots

- Before
Screen Shot 2021-03-21 at 1 17 50 PM
- After
Screen Shot 2021-03-21 at 1 18 02 PM

## Tests

- Check to make sure that for a class with sections taught in 2021 and before 2021, when filtering by Spring 21 the "last taught" section doesn't appear on the cards but it does when filtering by All

## Merging the PR

- Who should merge this PR?
  - [] The last reviewer to approve can merge it
  - [] preserve the history

## Add your changes to "Next Update" on the [release history page](https://github.com/thecourseforum/theCourseForum2/wiki/Release-History) once your PR gets merged!


tcf_website/templates/course/course.html
```diff
@@ -51,7 +51,7 @@ <h4 class="card-title">
                 {% for i in instructors %}
                     <li class="instructor {% if i.semester_last_taught != latest_semester %}old{% endif %}">
                         {% url 'course_instructor' course_id=course.pk instructor_id=i.pk as link %}
-                        {% include "common/rating_card.html" with title=i.full_name link=link rating=i.rating difficulty=i.difficulty gpa=i.gpa last_taught=i.semester_last_taught %}
+                        {% include "common/rating_card.html" with title=i.full_name link=link rating=i.rating difficulty=i.difficulty gpa=i.gpa last_taught=i.semester_last_taught current_semester=latest_semester %}
```
Bartleby2718, line 54: Seems like this wasn't deleted from past commits?

khushic: I went ahead and removed it, not sure if that's what you meant

Bartleby2718: It was! It's not supposed to be there, right?

khushic: Ahh yeah you're right it was from a previous edit that I deleted. Thanks for the catch!

andrewmniu:
Do you think it would be a good idea to maybe have the semester they last taught prior to the current one?

alex-shen1:
^ good suggestion! If you don't think that's necessary, I'd recommend at least spacing out the other information evenly across the now-expanded horizontal space to avoid having a chunk of empty whitespace where there used to be stuff.

khushic:
I think having the semester they taught prior to the current one could be confusing, I'll change the formatting so it's more spaced out

n4mahajan:
> I think having the semester they taught prior to the current one could be confusing, I'll change the formatting so it's more spaced out

Agreed. I also don't find that this is super necessary info that we need to provide anyway tbh unless someone wanted to see the pattern in which an instructor teaches a certain course (so that they can try and predict when they will teach it next again), but at that point they would need a longer history rather than just the previous time they thought. I'm of the thought that removing the redundancy as Khushi did is very solid

j-alicia-long:
Hmmm I feel like the way it originally was (displaying semester taught) signals information to the user that the button worked properly, so I'm more in favor of keeping it? We can vote on it at the meeting though.

alex-shen1:
General consensus: leave it as is

PR #463, opened 2021-03-24
Course Description on course page

### LINKED ISSUE #462: Add Description to Course Page
As discussed in the new feature proposal channel, it would be a good idea to include the course information not only on browse but on the actual course page.

borjainglefernandez:
## Status

- Done

## GitHub Issues addressed

- This PR closes #462 

## What I did

-Added the course description to the course page

## Screenshots

- Before
Screenshot 2021-03-24 at 10 24 35 AM
- After
Screenshot 2021-03-24 at 10 25 19 AM

## Tests

- Go to a couple course pages and ensure that the course description is there and is accurate

- Could look better? Maybe wrapping it in tCF blue or adjusting the font size or something

## Merging the PR

- Who should merge this PR?
  - [] I will merge it myself
  - [] The last reviewer to approve can merge it
  - [] preserve the history
  - [] squash-merge


andrewmniu (approved):
Looks good to me! Pretty simple change, so there's not much to say.

alex-shen1 (changes requested):
Like Andrew said, there isn't really a whole lot to say here code-wise, but I think you might want to get a basic mockup or something, because I don't think it's a great idea to have a relatively big chunk of text that's just hanging out in a `<p>` tag. At the very least, I'd recommend using a Bootstrap card to differentiate it from the other content, but there are other people on the team way better at this than me.

borjainglefernandez:
> Like Andrew said, there isn't really a whole lot to say here code-wise, but I think you might want to get a basic mockup or something, because I don't think it's a great idea to have a relatively big chunk of text that's just hanging out in a `<p>` tag. At the very least, I'd recommend using a Bootstrap card to differentiate it from the other content, but there are other people on the team way better at this than me.

Yes I agree, even just a div would help out a lot. 
I added a card and now it looks like this:
Screenshot 2021-03-25 at 9 49 32 AM
IMO it looks worse but I'll ask the design team for input if they have any.

tcf_website/static/course/course.css
```diff
@@ -0,0 +1,5 @@
+#course-eval-link{
```
borjainglefernandez, line 1: Is it better to go by class name instead of id? I noticed for most css files we go by class name.

borjainglefernandez:
image
image

tcf_website/static/course/course.css
j-alicia-long: class name is best used for styles that you need to apply to multiple similar elements!

For a one-off thing we usually recommend bootstrap classes instead of custom css. Do you think you can modify this?

j-alicia-long: class name is best used for styles that you need to apply to multiple similar elements!

For a one-off thing we usually recommend bootstrap classes instead of custom css. Do you think you can modify this?

j-alicia-long (changes requested):

kimberlyvo21:
Course Description Mockup

i made a mockup option for the description, I thought since we don't use the card headers or that specific type of card anywhere else it might be better to stay consistent and just use a plain card with bolded words for titles. I also added a place for pre-req so we can format it like the other cards on the page, also might be easier to identify certain necessities for classes. Idk if that info is easy to extract from the description but if it can't be done no biggie. Other than that i kept everything else pretty much the same.

borjainglefernandez:
Screenshot 2021-04-24 at 11 36 13 AM
Screenshot 2021-04-24 at 11 35 59 AM
 I tried my best to extract the pre-reqs from the course description itself but there are too many formats the pre-requisite(s) can come in that it just is not feasible. I believe @alex-shen1 put in a FOIA request for the pre-reqs so I'll add those in when the time comes.

borjainglefernandez:
Basic functionality down just need some design input:
Screenshot 2021-04-29 at 11 12 57 PM
Screenshot 2021-04-29 at 11 13 03 PM
This is meant to show what a class without a pre-req would look like and a class with a pre-req.

tcf_website/models/models.py
```diff
@@ -379,6 +381,31 @@ class Course(models.Model):
     def __str__(self):
         return f"{self.subdepartment.mnemonic} {self.number} | {self.title}"
 
+    def course_description_without_prerequisite(self):
+        """Returns course description without its pre-requisite(s)."""
+        return self.description.replace(self.pre_req, "")
+
+    def prerequisite(self):
```
borjainglefernandez, line 388: There is most definitely a less line heavy way of implementing this: open to suggestions.

tcf_website/models/models.py
```diff
@@ -369,6 +369,8 @@ class Course(models.Model):
     # Course number. Required.
     number = models.IntegerField(
         validators=[MinValueValidator(0), MaxValueValidator(99999)])
+    # Prerequisite. Optional.
+    pre_req = models.TextField(blank=True)
```
borjainglefernandez, line 373: Added this so it is easier for me to refer to it in `course_description_without_prerequisite`

borjainglefernandez: Also shaky about my use of `pre_req` and `prerequisite` I should probably standardize that right?

tcf_website/models/models.py
```diff
@@ -379,6 +381,31 @@ class Course(models.Model):
...
+        """Returns course pre-requisite(s) and assigns them to a model attribute pre-req."""
+        if self.description.find("Prerequisite: ") >= 0:
+            pre_req_to_end = self.description[self.description.find("Prerequisite: "):]
+            if pre_req_to_end.find(".") >= 0:  # In line pre-reqs
+                self.pre_req = pre_req_to_end[:pre_req_to_end.find(".") + 1]
+                return self.pre_req[len("Prerequisite: "):-1]
+            # At the end pre-reqs
+            self.pre_req = pre_req_to_end
+            return self.pre_req[len("Prerequisite: "):]
+        if self.description.find("Prerequisites: ") >= 0:
```
alex-shen1, line 398: Given that `"Prerequisites: "` is mostly the same as `"Prerequisite: "`, I'd suggest simply searching for `Prerequisite`, selecting the substring after that, and then dropping everything before/including the first colon. 
That way you don't need this extra if block.

tcf_website/models/models.py
```diff
@@ -379,6 +381,31 @@ class Course(models.Model):
...
+        """Returns course description without its pre-requisite(s)."""
+        return self.description.replace(self.pre_req, "")
+
+    def prerequisite(self):
+        """Returns course pre-requisite(s) and assigns them to a model attribute pre-req."""
+        if self.description.find("Prerequisite: ") >= 0:
+            pre_req_to_end = self.description[self.description.find("Prerequisite: "):]
+            if pre_req_to_end.find(".") >= 0:  # In line pre-reqs
+                self.pre_req = pre_req_to_end[:pre_req_to_end.find(".") + 1]
+                return self.pre_req[len("Prerequisite: "):-1]
```
alex-shen1, line 394: Why store the data as `"Prerequisite: {whatever}."` when you drop the leading `Prerequisite:` string and the period upon the return? Also applies to the other sections with similar logic.

tcf_website/templates/course/course.html
```diff
@@ -26,17 +26,38 @@ <h1 class="mr-md-3 text-nowrap">{{ course.code }}</h1>
...
-               title="Requires NetBadge Login"
-               class="text-tcf-blue font-weight-bold"
-               target="_blank"
-               href={{ course.eval_link }}>
-                Official UVA Student Evaluations
-                <i class="fa fa-external-link" aria-hidden="true"></i>
-            </a>
+            <!-- Course Description -->
+            {% if course.description|length > 0 %}
+            <div class="card-body">
```
alex-shen1, line 31: Not too sure on this one, but you might have to wrap this in a `class="card"` div to get the card background to show up properly.

alex-shen1 (changes requested):
Good work! Have a few overarching notes that don't attach to any particular code segment:
- In case you or anyone else was wondering, I did a little investigating into how many courses this methodology is getting, out of the total number:
image
It seems that there are actually 3030 `Course` instances with "prerequisite" somewhere in the description (`icontains` is non-case-sensitive):
image
- More broadly, I think the logic in your `prerequisite()` function should be added to the initial `load_semester` script we run each semester. This is because otherwise the function will be recomputing and updating the database each and every time a course page is opened, which we probably don't want for performance reasons. Alternatively, if we're going to be recomputing the string every time, there's no point in saving it in the database because as far as I can tell, the templates only call the functions. IIRC I updated `load_semester` a while back to be able to be run multiple times without impacting already-existing data, so basically we could just compute the prerequisite string once, save it in the DB, and then just read it as a field.
- It might be good to have each course in the prereqs string hyperlinked to the actual course page, but that seems technically challenging and perhaps outside the scope of this PR. Along with the other stuff, this might take some time, but I think the course description is already good to go so we might want to merge that portion and branch off in another PR.

tcf_website/models/models.py
borjainglefernandez: Ah very smart move

tcf_website/models/models.py
borjainglefernandez: Yeah I wrote this in a hurry, good catch

tcf_website/models/models.py
```diff
@@ -379,6 +381,31 @@ class Course(models.Model):
     def __str__(self):
         return f"{self.subdepartment.mnemonic} {self.number} | {self.title}"
 
+    def course_description_without_prerequisite(self):
+        """Returns course description without its pre-requisite(s)."""
+        return self.description.replace(self.pre_req, "")
+
+    def prerequisite(self):
+        """Returns course pre-requisite(s) and assigns them to a model attribute pre-req."""
+        if self.description.find("Prerequisite: ") >= 0:
```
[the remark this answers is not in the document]
borjainglefernandez: oh yeah i forgot that was a python thing

borjainglefernandez:
> Good work! Have a few overarching notes that don't attach to any particular code segment:
> 
> * In case you or anyone else was wondering, I did a little investigating into how many courses this methodology is getting, out of the total number:
>   image
>   It seems that there are actually 3030 `Course` instances with "prerequisite" somewhere in the description (`icontains` is non-case-sensitive):
>   image
> * More broadly, I think the logic in your `prerequisite()` function should be added to the initial `load_semester` script we run each semester. This is because otherwise the function will be recomputing and updating the database each and every time a course page is opened, which we probably don't want for performance reasons. Alternatively, if we're going to be recomputing the string every time, there's no point in saving it in the database because as far as I can tell, the templates only call the functions. IIRC I updated `load_semester` a while back to be able to be run multiple times without impacting already-existing data, so basically we could just compute the prerequisite string once, save it in the DB, and then just read it as a field.
> * It might be good to have each course in the prereqs string hyperlinked to the actual course page, but that seems technically challenging and perhaps outside the scope of this PR. Along with the other stuff, this might take some time, but I think the course description is already good to go so we might want to merge that portion and branch off in another PR.

I think the idea of doing "Prerequisite" and then finding the first ":" should work for most classes and ensure we do not have invalid prerequisites such as: 
image

borjainglefernandez:
Screenshot 2021-09-18 at 11 18 45 PM
Screenshot 2021-09-18 at 11 18 52 PM

Should be good to go now to re-review.

tcf_website/models/models.py
```diff
@@ -379,6 +381,34 @@ class Course(models.Model):
...
+            # Get rid of title of "Prerequisite"
+            pre_req_no_title = from_pre_req_to_end[from_pre_req_to_end.find(":") + 1:]
+            # Check if in-line or not
+            if pre_req_no_title.find(".") > 0:
+                self.pre_req = pre_req_no_title[:pre_req_no_title.find(".")]
+            else:
+                self.pre_req = pre_req_no_title
+        else:
+            self.pre_req = ""
+        return self.pre_req
```
vn6, line 410: Looks like there's a lot of repeated code between these two functions. It might be better to combine the functionality here into a common function like `parse_course_description` which returns a tuple `(course_desc_without_prereqs, prereqs)`.  

Then, `course_description_without_pre_req`  and `compute_pre_req` can just call `parse_course_description` and return the proper values.

borjainglefernandez: Great idea!

tcf_website/models/models.py
```diff
@@ -379,6 +379,41 @@ class Course(models.Model):
...
+
+            # Check if in-line or not for pre_req
+            if pre_req_no_title.find(".") > 0:
+                pre_req = pre_req_no_title[:pre_req_no_title.find(".")]
+            else:
+                pre_req = pre_req_no_title
+
+            # Check whether it is inline or at end for extracting course_description_without_pre_req
+            if from_pre_req_to_end.find(".") > 0:
+                course_description_without_pre_req = self.description.replace(
```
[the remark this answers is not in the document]
borjainglefernandez: It does but then I would have to recompute the indices as opposed to having them implicitly computed for me.

alex-shen1 (approved):

Issue #412, opened 2021-02-22
Dropdown menu for courses


alex-shen1:
**Is your feature request related to a problem? Please describe.**
Basically how Lou's list does it, with each school as its own section that can be hidden. CLAS should be open first, also like Lou's.

**Describe the solution you'd like**
Would require some reworking of the data to "promote" everything currently in "Other Schools" to actual schools.

Proposal/mockup slide is [here](https://docs.google.com/presentation/d/1lTuTiClBFfQnKL0okLd7OGjzJbBMhIr6oX4PpVrZGtQ/edit#slide=id.gbd2773688b_0_0).

PR #465, opened 2021-03-27
Course dropdown

jackliu612:
## Status

- "Done" with more styling changes possible based off of reviews

## GitHub Issues addressed

- #412 

## What I did

- Changed the main browse menu to have expanding accordions instead of listing everything directly

## Screenshots

- Before
image

- After
image

## Tests

- No automated tests done, but you can go to main browse and check that dropdowns work as expected

- Original mock up has up/down arrows depending on whether a section is expanded but I'm not sure how to implement that well. Also there was possible talk about alternating heading colors to add some contrast

## Merging the PR

- Who should merge this PR?
  - [x] I will merge it myself
  - [ ] The last reviewer to approve can merge it
  - [ ] preserve the history
  - [x] squash-merge


Bartleby2718:
Just a thought: would it be possible to cache which dropdowns were expanded? For example, a nursing student will have to scroll all the way to the bottom and expand `School of Nursing` every time they visit the browse page.

tcf_website/templates/browse/browse.html
```diff
@@ -4,29 +4,20 @@
...
 {% endblock %}
 
 {% block content %}
-    {% comment %} <div class="container">
-        {% include "../common/notification.html" %}
-    </div> {% endcomment %}
-    <div class="browse container">
-        <!-- Google adsense -->
-        <div class="container mb-4">
-            {% include "../common/banner_ad.html" %}
```
borjainglefernandez, line 17: Don't we want to keep the ads here or are we moving them somewhere else?

borjainglefernandez (commented):
Overall, you did a fantastic job and I really like this feature. I tested it both on desktop and mobile and it works perfectly well. A few comments that could enhance this feature slightly:
-Could maybe make the arrows point up when they are not toggled to show which departments are being shown
-The inclusion of a "View all" button and a "Collapse all" button
-Saving information about what departments are toggled when the user leaves the browse page so that when the user comes back the same departments that were open stay open and the same departments that were closed stay closed (This may be way more effort than it's worth but just throwing it out there)
-As a matter of personal preference, I think it might be better to leave all of the departments collapsed at first or only showing College of Arts & Sciences.
-Was the ad banner removed on purpose?
-I think this might be impossible to address since you're using bootstrap elements if I am not mistaken but when collapsing a department, there is a little bit of lag right before the department closes
I really do not view any of these suggestions as critical, which is why I am not requesting changes but they may be something to consider.

n4mahajan:
> Overall, you did a fantastic job and I really like this feature. I tested it both on desktop and mobile and it works perfectly well. A few comments that could enhance this feature slightly:
> -Could maybe make the arrows point up when they are not toggled to show which departments are being shown
> -The inclusion of a "View all" button and a "Collapse all" button
> -Saving information about what departments are toggled when the user leaves the browse page so that when the user comes back the same departments that were open stay open and the same departments that were closed stay closed (This may be way more effort than it's worth but just throwing it out there)
> -As a matter of personal preference, I think it might be better to leave all of the departments collapsed at first or only showing College of Arts & Sciences.
> -Was the ad banner removed on purpose?
> -I think this might be impossible to address since you're using bootstrap elements if I am not mistaken but when collapsing a department, there is a little bit of lag right before the department closes
> -Might want to move more popular schools like McIntire closer to the top since we are not sorting alphabetically
> I really do not view any of these suggestions as critical, which is why I am not requesting changes but they may be something to consider.

Definitely agree about the arrow point and making school of arts/sciences the default option open (maybe both the college and eschool but that might look ugly). Also, I think Jack just preserved the original order of the schools so rearranging them might be out of the scope of this PR? View all/collapse all is also a good suggestion but can be left on the back-burner tbh. Would also be nice to save the dropdown preferences but don't know how difficult it is. These are all solid suggestions though.

j-alicia-long:
> Just a thought: would it be possible to cache which dropdowns were expanded? For example, a nursing student will have to scroll all the way to the bottom and expand `School of Nursing` every time they visit the browse page.

^Agree that caching/rearranging is out of scope for this PR, but I'd like to see a "favorite" feature that pins sections at the top

j-alicia-long:
btw I think this is fine to merge as-is, and we can make new issues for the caching/favoriting features. @jackliu612 once you resolve the merge conflict I can approve!

alex-shen1:
SMH all you engineers really just completely overlooked that he accidentally deleted the College from our site lmao

alex-shen1 (approved):
Compiled various discussions into #487, so we're good to go! Incidentally, since CLAS is so big the actual browse page doesn't look any different on my monitor at 100% zoom lol:
image

PR #503, opened 2021-05-18
Warning for user if they submit 0 total hours/week

### LINKED ISSUE #502: 0 Hours/Week Warning
Warn the user if they have entered 0 total hours of work per week in their review. Entering any hours is not required but it's highly improbable the user truly had 0 hours of work per week in the class (i.e. some reviews in CS 2150 of hours/week). Warning only appears once as it might actually be 0, but hopefully this will prompt them to enter the hours.

n4mahajan:
## Status

- Done? (Subject to changes)

## GitHub Issues addressed

- This PR closes #502

## What I did

- Added a warning modal for users if they enter 0 *total* hours/week in their review.
- The modal only appears once, the first time after a user clicks the submit button, if their review has 0 hours. 
- The Ajax request that checks if the review has 0 total hours is nested in the initial Ajax request that checks if the review is a duplicate.
- A method was added in the Views file for review to handle checking if the review has 0 hours. 

## Screenshots

- Before
- After

## Tests

- Write several reviews: reviews that are duplicates for the same course, reviews that have 0 hours/week but is not a duplicate review, reviews that are neither duplicates nor have 0 hours/week, etc. Ensure that when writing a review, the zero hours/week modal only appears once when the user clicks the submit button and the review has zero hours/week. Even if the user clicks submit again and the review still has zero hours/week, the review should submit and no modal should appear. In all cases, if a review is a duplicate, the duplicate review modal should appear before the zero hours/week modal; the zero hours/week modal should still show if it didn't because of the duplicate review modal. 

- Not sure if we want to make the requirements more strict for entering the hours/week. 
- Also not sure if we want to tackle this issue differently when prompting users to properly enter the hours/week, since a lot of people end up foregoing entering it. 
- I wrote the modal rather fast. It could probably be written better so I'm more than open to suggestions about how to make the modal better.
- There's also the edge case where the user enters 0 hours/week for a course and the modal appears, but then they change the course they're reviewing in which the modal no longer appears (if needed). Not sure how we want to handle this edge case but I think it's rather negligible. 

## Merging the PR

- Who should merge this PR?
  - [] I will merge it myself
  - [x] preserve the history
  - [] squash-merge


tcf_website/templates/reviews/new_review.html
```diff
@@ -34,6 +35,8 @@ <h1 class="h2 mt-2 mb-4">Write a review</h1>
     $('[data-toggle="tooltip"]').tooltip()
   });
 
+  // Counter to keep track of number of times "zero-hours-review" modal is shown
+  var count = 0;
```
alex-shen1, line 39: Since the only values `count` ever takes on are `0` and `1`, it might be better to make it a boolean?

tcf_website/templates/reviews/zero_hours_modal.html
```diff
@@ -0,0 +1,21 @@
...
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title">Zero Hours per Week in Review</h5>
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+          It looks like you submitted zero total hours of work per week for this class. Please double check
```
alex-shen1, line 12: Minor grammar nitpicks (lmao), feel free to override: I _think_ `double-check` is a hyphenated compound word - could be wrong on that one though. Also I'm pretty sure there should be a comma between `work` and `as`?

tcf_website/templates/reviews/new_review.html
```diff
@@ -48,13 +51,28 @@ <h1 class="h2 mt-2 mb-4">Write a review</h1>
...
+                            // Only show "zero-hours-review" modal once
+                            if(check.zero && count == 0) {
+                                $("#zero-hours-review").modal();
+                                count = 1;
+                            } else {
+                                // Timeout button for 3 seconds so user can't spam button while form is submitting
+                                $("#submitBtn").prop("disabled", true);
+                                setTimeout(enableButton, 3000);
+                                // $("#reviewForm") didn't work for some reason
+                                // Not a duplicate review, so proceed with normal form submission for new review
```
alex-shen1, line 70: Comment should be updated to reflect how at this point not only is the review not a duplicate, but that the user has been shown 0 hours modal

alex-shen1 (commented):
Functionality works great! Just a few minor comments and this should be good to go.

tcf_website/templates/reviews/new_review.html
n4mahajan: hmmm yeah good call, would be more intuitive

tcf_website/templates/reviews/zero_hours_modal.html
n4mahajan: From what I can tell either is fine I think? I'll change it to double-check to be safe though. Should be a comma between "work" and "as" like you mentioned though

tcf_website/templates/reviews/new_review.html
n4mahajan: ^ has been shown it already or it was never necessary to show in the first place

alex-shen1 (approved):

PR #531, opened 2021-10-31
Revert "Revert "Save Course""

alex-shen1:
Reverts thecourseforum/theCourseForum2#527

lmao this is kinda scuffed but basically we wanted to remove save course from dev while preserving its changes in a branch


alex-shen1:
We still don't have local login figured out so not entirely sure how this could be tested locally, but I have a feeling that having a migration file committed _might_ fix the issue?

PR #552, opened 2021-11-21
Update files to use new .env variable

### LINKED ISSUE #519: Consolidate the Elasticsearch environmental variables
Right now, we use 4 separate environmental variables for configuring Elasticsearch endpoints (the redacted portion is the host URL, which we don't want getting leaked as this is a public repo, check the [secrets repo](https://github.com/thecourseforum/tCF-Secrets) to see the full files for yourself):

image

Notice that these values are essentially identical all the way up to `/engines/`, and even then there are 2 duplicates for engine name. This is pretty redundant, so we can simply reduce it down to one environmental variable and then augment it with whatever we need at runtime. For example, you could change `.env` to have `ELASTICSEARCH_ENDPOINT={HOST}/api/as/v1/engines/` (fill in `{HOST}` with the real value, which I can't type here) and then when you need to use the endpoint in `search.py`, you set `course_search_endpoint = ELASTICSEARCH_ENDPOINT + 'tcf-courses/search'`.

This ticket can be broken down into several tasks:
- [ ] Change `.env` to consolidate the variables and create a PR to put these changes in [the secrets repo](https://github.com/thecourseforum/tCF-Secrets)
- [ ] Update logic wherever our Elasticsearch endpoints are used by a Python file (i.e. `search.py` in views and the management command `index_elasticsearch.py`) to augment the one endpoint variable as previously outlined
- [ ] Update config files that use these environmental variables e.g. `/config/.env.example` and the `.yml` files in `/.github/workflows`
- [ ] Get someone with permission to edit repo secrets (pretty sure it's set up so that only exec can) to update GitHub Secrets and our Heroku config to the appropriate new values

I get that this wall of text is probably pretty intimidating, but it shouldn't be too technically challenging (most of it is just making small text edits), and it's a great way to get an understanding of how our environments work!

obs4qqf:
## Status

- Done

## GitHub Issues addressed

- This PR closes #519 

## What I did

- Updated search.py and index_elasticsearch.py endpoint variables to use 1 .env variable and then modify endpoint at runtime
- Updated .env.example, app.yaml.template, .yml files in /.github/workflows, and base.py in /tcf_core to reduce used .env variables

## Screenshots

- Before (to show search still works)
before1
before2

- After
after1
after2

## Tests

- Checked that course and instructor searches still work locally

- The last part of the linked issue needs to be addressed: "Get someone with permission to edit repo secrets (pretty sure it's set up so that only exec can) to update GitHub Secrets and our Heroku config to the appropriate new values"
- This PR is also linked to a PR in the secrets repo that updates the .env variables
- Some files, like base.py and app.yaml.template, were changed to ensure the search worked, but these weren't mentioned in the original issue checklist - is this fine?

## Merging the PR

- Who should merge this PR?
  - [] I will merge it myself
  - [x] preserve the history
  - [] squash-merge


alex-shen1 (approved):
Looks great! Note to exec: need to edit Heroku/GitHub secrets before this can get merged thoguh

PR #561, opened 2022-01-30
Review autofill from course instructor page

### LINKED ISSUE #156: auto-populate write a review form
when clicking on write a review option on course page autofill known info:
- course prefix/number
- professor
- semester

akh4rf:
## Status

- Done

## GitHub Issues addressed

- This PR closes #156

## What I did

- Created new path to the `new_review()` view in order to pass query params (subdepartment_id, course_id, instructor_id) needed to autofill the review form
- Updated the `course_professor.html` template to utilize this new url path when the "Add your review!" button is pressed
- Updated the jQuery in `new_review.js` that populated the select field options in the form, auto-selecting the options matching the query params (if applicable)

## Screenshots

- Before

image

- After (note the url path)

image

## Tests

- After pull, navigate to any course instructor page and click the "Add your review!" button. The review form should have the Subject, Course, and Instructor fields autofilled.

- Did not autofill semester, as people are allowed to review instructors for classes taught in past semesters.

## Merging the PR

- Who should merge this PR?
  - [] I will merge it myself
  - [] preserve the history
  - [] squash-merge

### DIFF
--- a/tcf_website/urls.py
+++ b/tcf_website/urls.py
@@ -1,4 +1,5 @@
 """Routes URLs to views"""
+# pylint: disable=line-too-long
 
 from django.urls import include, path
 
@@ -20,6 +21,10 @@ urlpatterns = [
     path('instructor/<int:instructor_id>',
          views.instructor_view, name='instructor'),
     path('reviews/new', views.new_review, name='new_review'),
+    path(
+        'reviews/new/subdept/<int:subdepartment_id>/course/<int:course_id>/instr/<int:instructor_id>',
+        views.new_review,
+        name='new_review_auto'),
     path('reviews/<int:pk>/delete',
          views.DeleteReview.as_view(), name='delete_review'),
     path('reviews/<int:review_id>/edit',
--- a/tcf_website/views/review.py
+++ b/tcf_website/views/review.py
@@ -62,8 +62,9 @@ def downvote(request, review_id):
 
 
 @login_required
-def new_review(request):
+def new_review(request, subdepartment_id=None, course_id=None, instructor_id=None):
     """Review creation view."""
+    print(subdepartment_id, course_id, instructor_id)
 
     # Collect form data into Review model instance.
     if request.method == 'POST':
--- a/tcf_website/static/reviews/new_review.js
+++ b/tcf_website/static/reviews/new_review.js
@@ -13,6 +13,21 @@ jQuery(function($) {
     clearDropdown("#instructor");
     clearDropdown("#semester");
 
+    // If coming from sidebar:
+    // Path will be /reviews/new
+    //    => path = ['', 'reviews', 'new']
+
+    // If coming from course instructor page (autofill):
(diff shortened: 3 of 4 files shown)


tcf_website/urls.py
```diff
@@ -20,6 +21,10 @@
     path('instructor/<int:instructor_id>',
          views.instructor_view, name='instructor'),
     path('reviews/new', views.new_review, name='new_review'),
+    path(
+        'reviews/new/subdept/<int:subdepartment_id>/course/<int:course_id>/instr/<int:instructor_id>',
```
alex-shen1, line 25: Since you're already using the names `subdept`, `instr` in the path, what do you think of modifying the URL parameter names like`<int:subdepartment_id>` to be consistent? Maybe the abbreviation will shorten the line so Pylint no longer complains.

On the other hand, it does kinda break with the naming conventions established by the actual models e.g. `Course.subdepartment` so maybe not.

tcf_website/views/review.py
```diff
@@ -62,8 +62,9 @@ def downvote(request, review_id):
 
 
 @login_required
-def new_review(request):
+def new_review(request, subdepartment_id=None, course_id=None, instructor_id=None):
     """Review creation view."""
+    print(subdepartment_id, course_id, instructor_id)
```
alex-shen1, line 67: Remove print!

alex-shen1 (changes requested):
Haven't been able to test thoroughly yet, but small comments

tcf_website/views/review.py
```diff
@@ -62,7 +62,7 @@ def downvote(request, review_id):
 
 
 @login_required
-def new_review(request):
+def new_review(request, subdept_id=None, course_id=None, instr_id=None):
```
alex-shen1, line 65: To continue our discussion re: encoding data in URL path vs. GET parameters: I'm of the opinion that there should be a one-to-one mapping of URL paths to "pages". Granted, a lot of pages on the site (course, course-instructor, etc.) all have similar formats and are differentiated only by the path so I can see an argument there, but it feels like the review form is just one thing regardless of URL - you wouldn't say "this is the page where you can review CS 2150 and only CS 2150", if that makes sense.

Also, I'm assuming that defining fields in the URL path requires this view here to have parameters that correspond to the names defined in `urls.py`, that don't get used? Considering that all the logic is done in JS from simply accessing the URL, it feels unnecessary to involve the view at all, but if my assumption is right then we're forced to do so.

Would it be significantly difficult to convert this to a GET format? My instinct is no given that you've already figured out the URL path way, but hey I also never managed to figure this out ¯\\\_(ツ)_/¯ Ultimately, it's great that we have this now and it's your call - just wanted to give my take.

akh4rf: Can definitely implement it using query string params! Updating now!

alex-shen1 (approved):
I had a comment about changing the name of `path` to `params` in the JS file now that technically you're now accessing the GET parameters instead of the URL path, but since CI didn't run last time I figured I'd just do it myself to see if a new commit would make CI work this time. Turns out, it did!

PR #574, opened 2022-02-27
Remove makemigrations from deployment scripts

alex-shen1:
As of #571, migrations are required to be committed directly into the repo for...version control reasons. As such, we should not be running `makemigrations` in the deployment scripts to ensure that we don't forget.


jackliu612 (approved):
LGTM 👍

PR #587, opened 2022-03-23
Precommit hook updates

alex-shen1:
Also need to update the script to automatically stage autoformat changes before opening this PR


alex-shen1:
All this PR does is remove the color characters (because while they look nice in certain terminals, they're just spam characters in others), update the script to automatically add files again after autoformatting, and update the manual precommit script to match the automated ones (but it shouldn't be necessary to run anymore)

jackliu612 (approved):
Lgtm

PR #588, opened 2022-04-03
Sorting instructors on course detail page

### LINKED ISSUE #264: Sorting Instructors by Rating/Difficulty/GPA in Course Detail
On the [course detail page](https://thecourseforum.com/course/11164), have the option to sort instructors by rating, difficulty, and GPA (similar to sorting the courses this way). By default, instructors should be sorted by how recently they have taught a class (originally mentioned in #424). Here's an example of how the list is currently unordered:

image

obs4qqf:
## Status

- Done

## GitHub Issues addressed

- This PR closes #264 

## What I did

- Reused sorting toolbar functionality from departments page with some modifications (ex. added "Last Taught" sort button)
- Updated toolbar.html component to include either sorting by last taught or by course ID based on whether the component is used in the departments versus course detail page
- Added functions for sorting by last taught to sorting.js and sortClasses.js
- Moved some static files under /common for reusability

## Screenshots

- Before
before

- After (to show that sorting works)
- Sort by last taught works (default sorting
recent

- Sort by rating works
rating

- Sort by difficulty works
diff

- Sort by gpa works
gpa

## Tests

- Tested that various sorting buttons worked (as seen above)
- Tested random sequences of sorting buttons to verify sorting works for edge cases
- Verified that sorting by last taught correctly sorts courses with the same year but different session

- I reorganized some static files because of the reused functionality, but I'm not sure if the way I did this is confusing

## Merging the PR

- Who should merge this PR?
  - [] I will merge it myself
  - [] preserve the history
  - [x] squash-merge


tcf_website/static/common/bind-sort.js
```diff
@@ -0,0 +1,12 @@
+import { sortByNumber, sortByRating, sortByDifficulty, sortByGpa, sortByRecency, sortByRecencyOnce } from "./sortClasses.js";
+
+if ($("#number-sort-btn").length) {
```
alex-shen1, line 3: Can you add a comment here explaining that these `if`s need to be here because `toolbar.html` is reused across pages and we need different sorting buttons on dept/course?

tcf_website/static/common/sortClasses.js
```diff
@@ -5,6 +5,7 @@ let sortNumberAsc = 1;
 let sortRatingAsc = -1;
 let sortDifficultyAsc = 1;
 let sortGpaAsc = -1;
+let sortRecencyAsc = -1;
```
alex-shen1, line 8: Kinda out of scope for your PR because you were following the existing pattern, but do you know of any reason why these can't be bools? lol

tcf_website/static/common/sortClasses.js
```diff
@@ -37,10 +38,16 @@ const sortByNumber = () => {
 const sortByRating = () => {
     activeSort = "rating";
     if (!$("#rating-sort-btn").hasClass("active")) {
-        $("#number-sort-btn").removeClass("active");
+        if ($("#number-sort-btn").length) {
```
alex-shen1, line 41: Same thing as before - I assume these are just proxies for checking whether or not the page that's using the sort function has these fields?

tcf_website/static/common/sortClasses.js
```diff
@@ -83,6 +102,37 @@ const sortByGpa = () => {
     selectOrderOfSort(sortGpaAsc, "#gpa-sort-btn", "GPA");
 };
 
+const sortByRecency = () => {
```
alex-shen1, line 105: And then add a comment here saying that this doesn't need those checks because it's only used on 1 page

tcf_website/static/common/sortClasses.js
```diff
@@ -83,6 +102,37 @@ const sortByGpa = () => {
...
+        $("#diff-sort-btn").html("Difficulty");
+        $("#gpa-sort-btn").html("GPA");
+        $("#recency-sort-btn").addClass("active");
+    } else {
+        sortRecencyAsc *= -1;
+    }
+    selectOrderOfSort(sortRecencyAsc, "#recency-sort-btn", "Last Taught");
+};
+
+const sortByRecencyOnce = () => {
```
alex-shen1, line 121: This function is essentially a copy of `sortByRecency()` - there's probably a way to make the code more concise? One way I can think of is defining this as a two-line function that first sets `sortRecencyAsc = 1` and then immediately calls `sortByRecency()`, which essentially accomplishes the same thing, but that feels kinda scuffed.

tcf_website/static/common/sorting.js
```diff
@@ -20,6 +20,38 @@ function cmpByProp(prop, asc) {
         if (prop === "date") { // convert date string to int
             valA = dateToInt(valA);
             valB = dateToInt(valB);
+        } else if (prop === "recency") {
+            try {
```
alex-shen1, line 24: Since you use the same regex parsing code for `valA` and `valB`, I'd define a helper function you can call twice so you don't repeat code.

tcf_website/static/common/sorting.js
```diff
@@ -20,6 +20,38 @@ function cmpByProp(prop, asc) {
         if (prop === "date") { // convert date string to int
             valA = dateToInt(valA);
             valB = dateToInt(valB);
+        } else if (prop === "recency") {
+            try {
+                const sessionA = valA.match(/^[a-zA-Z]+/)[0];
+                valA = valA.match(/-?[0-9]\d*(\.\d+)?/)[0];
```
alex-shen1, line 26: I'm like 90% sure that what you're doing here is setting `sessionA` to the season and `valA` to the year, but it might be more descriptive to first assign the year value to an intermediate variable (e.g. `yearA`) and then combine it all into `valA` at the end for readability.

tcf_website/static/common/sorting.js
```diff
@@ -20,6 +20,38 @@ function cmpByProp(prop, asc) {
...
+                const sessionA = valA.match(/^[a-zA-Z]+/)[0];
+                valA = valA.match(/-?[0-9]\d*(\.\d+)?/)[0];
+                // last taught session taken into account for sorting
+                if (sessionA === "January") {
+                    valA = parseFloat(valA + "0");
+                } else if (sessionA === "Spring") {
+                    valA = parseFloat(valA + "1");
+                } else if (sessionA === "Summer") {
+                    valA = parseFloat(valA + "2");
+                } else {
```
alex-shen1, line 34: I'm not sure if this is best practice or excessively obvious documentation, so feel free to ignore this if you disagree, but it might be helpful to make it explicit that the `else` here _should_ only ever be the Fall semester. Maybe even do another `if` and change the `else` to throw an error if we get unexpected data.

tcf_website/templates/course/course.html
```diff
@@ -17,7 +17,7 @@
             {% include "../common/leaderboard_ad.html" with ad_slot="9533576562" %}
         </div>
         <div class="container-lg">
-            {% include "common/toolbar.html" with breadcrumbs=breadcrumbs active_text=latest_semester hidden_text="All" %}
+            {% include "common/toolbar.html" with breadcrumbs=breadcrumbs active_text=latest_semester hidden_text="All" sorting=True page="courses" %}
```
alex-shen1, line 20: I think instead of using a context variable to track what page you're on, you can just get the URL directly. See [here](https://stackoverflow.com/questions/2882490/how-to-get-the-current-url-within-a-django-template) for details.

alex-shen1 (changes requested):
Functionally works great! Have some comments on implementation, but good job!

tcf_website/static/common/bind-sort.js
obs4qqf: Sounds good, I'll be sure to add that in.

tcf_website/static/common/sortClasses.js
obs4qqf: We might've talked about this earlier, but I think we used this in cmpByProp() to make directional sorting easier.

tcf_website/static/common/sortClasses.js
obs4qqf: Yup, just added it!

tcf_website/static/common/sortClasses.js
obs4qqf: I tried doing a two-line function, but I ran into problems with sortRecencyAsc not updating to -1 when the last taught sort button was not previously active. I ended up condensing the 2 sortByRecency functions into a single function, but this might've made the code more confusing (like in the way I updated bind-sort.js). I'm not sure if I should use this condensed function or leave the repeated code for readability purposes.

tcf_website/static/common/sorting.js
obs4qqf: Sounds good, I added getSessionNum() to fix this.

tcf_website/static/common/sorting.js
obs4qqf: Yep, I just fixed this within getSessionNum()

tcf_website/static/common/sorting.js
obs4qqf: I added another `if` to handle the Fall semester, and also changed the `else` to sort unexpected data to the end.

tcf_website/templates/course/course.html
obs4qqf: That makes sense- I updated `toolbar.html`, `department.html`, and `course.html` to reflect this change.

tcf_website/static/common/sortClasses.js
obs4qqf: Yeah, I added more comments to document this in case it was confusing.

alex-shen1 (changes requested):
It's been a long time, but one bit of design feedback that just occurred to me: 
image

There's quite a bit of blank space created with the current component layout. Would it be better to put both the course name/code on the same y-level as the sort buttons to reduce wasted vertical space?

obs4qqf:
I looked into adding everything on one line, and there might be some longer class titles that have less room to wrap around. I think having some empty space might keep the header from looking too busy, but I'm also open to anything. Is there one that looks better to you?
tcfscr
tcfscr2

PR #591, opened 2022-04-03
Added modal to new review button

### LINKED ISSUE #492: Review Sidebar Not Showing Login Modal
When you click "Review" on the sidebar, it goes straight to sign in with google instead of doing the login modal popup

ajnye:
## Status

- Done

## GitHub Issues addressed

- This PR closes #492

## What I did

- Changed the pathing for the login modal to use django variables for its id and path so that when called the redirect link afterwards can be set to whatever it is created with in the `{% include with %}` variables.

## Screenshots

Can't really upload screenshots, but for both "Review" on the sidebar and "Write a Review" on the landing page, it shows the modal then redirects to the new review after login. 
- Before
- After

## Tests

Load the changes and try to use both of the review buttons, logged in or not, to make sure they work properly of showing the modal, logging in, and redirecting to the new review page.

## Merging the PR

- Who should merge this PR?
  - [ ] I will merge it myself
  - [ ] preserve the history
  - [ ] squash-merge


alex-shen1 (changes requested):
Looks good functionally, but I believe there should be a way to write this more efficiently - the 2 templates you make for the review modal/button are essentially duplicates of the existing ones, but with 1-2 string changes. If you refactor the original modal/button to have these strings be context variables instead of being hardcoded, you can use the `with` keyword in the `{% include %}` to reuse the same template (see [this](https://stackoverflow.com/questions/11639306/assign-variables-to-child-template-in-include-tag-django) for details). 

As a side note, this is also inefficient in its own way because we still have essentially 2 copies of the HTML loaded in the actual page at runtime, even if we only have 1 in the source. A "best" solution here is _probably_ to change the trigger from a direct link to calling a JS function that dynamically changes the modal's property and then opens it.... but given that the difference is like ~20 lines of HTML code, if you don't want to do it that way I think the Django method is more than sufficient.

ajnye:
> Looks good functionally, but I believe there should be a way to write this more efficiently - the 2 templates you make for the review modal/button are essentially duplicates of the existing ones, but with 1-2 string changes. If you refactor the original modal/button to have these strings be context variables instead of being hardcoded, you can use the `with` keyword in the `{% include %}` to reuse the same template (see [this](https://stackoverflow.com/questions/11639306/assign-variables-to-child-template-in-include-tag-django) for details). Then I believe they become normal variables in the template which you can access with `{{ variable }}`.
> 
> As a side note, this is also inefficient in its own way because we still have essentially 2 copies of the HTML loaded in the actual page at runtime, even if we only have 1 in the source. A "best" solution here is _probably_ to change the trigger from a direct link to calling a JS function that dynamically changes the modal's property and then opens it.... but given that the difference is like ~20 lines of HTML code, if you don't want to do it that way I think the Django method is more than sufficient.

Used `with` to successfully make it use one copy instead of having two.

alex-shen1 (approved):

PR #594, opened 2022-04-14
Update review drive banner for Spring 2022

alex-shen1:


obs4qqf:
Just to clarify, is the banner meant to be added to the browse page only? Otherwise, it looks good on my end!

alex-shen1:
Yeah, that's what we've done in the past.

PR #595, opened 2022-04-25
Display Instructor/Course First

### LINKED ISSUE #167: Display only either instructor or course search results (no point in showing both for same query)

ajnye:
## Status

- Done

## GitHub Issues addressed

- This PR closes #167 #136 

## What I did

- Made z-scores on the top 10 results for instructors and courses, choosing which one has a higher z-score and is more likely to be relevant to the search. Which one is chosen is the tab selected first in the search engine.
- Since courses are more likely to be chosen and the z-score does not always choose the one that may be more relevant, when the letters searched are 4 or under, the courses are prioritized if there are some that are relevant to the search. Also prioritizes classes when only one result or none are found for both courses and instructors.

## Screenshots

- Before
Screen Shot 2022-04-25 at 12 33 02 PM
- After
Screen Shot 2022-04-25 at 12 32 51 PM

## Tests

- Anyone who tests this PR needs to try maybe different likely searches of students to see if it gives the desired result more often than not. When the code is unsure, it should prioritize courses, as courses are more relevant on average than specific professors.

- Does anyone know a better way to make the search more likely choose the correct one, or prioritize courses more often? Certain edge cases do not work very well, and the only reason 4 letters and under prioritizes courses is that it over-represents professors otherwise in the searhc.

## Merging the PR

- Who should merge this PR?
  - [ ] I will merge it myself
  - [ ] preserve the history
  - [ ] squash-merge


tcf_website/views/search.py
```diff
@@ -17,14 +18,61 @@ def search(request):
...
+    args = set_arguments(query, courses, instructors, courses_first)
     context_vars = args
 
     # Load template view
     return render(request, 'search/search.html', context_vars)
 
 
+def decide_order(query, courses, instructors):
+    """Decides if courses or instructors should be displayed first."""
+    # Returns True if courses should be prioritized, False if instructors should be prioritized
```
alex-shen1, line 33: This comment should probably go into the function docstring itself instead of being right underneath it

tcf_website/views/search.py
```diff
@@ -17,14 +18,61 @@ def search(request):
...
+    if len(courses['results']) > 1:
+        course_scores = [x['score'] for x in courses['results']]
+
+        courses_mean = statistics.mean(course_scores)
+
+        courses_stddev = statistics.stdev(course_scores, courses_mean)
+
+        if courses_stddev == 0:
+            courses_stddev = 1
+        courses_z = courses['results'][0].get('score') - courses_mean
```
alex-shen1, line 45: Are the search results guaranteed to be sorted descending by score? Intuitively it makes sense, just wondering if that's confirmed anywhere.

tcf_website/views/search.py
```diff
@@ -17,14 +18,61 @@ def search(request):
...
+        courses_z /= courses_stddev
+    elif len(courses['results']) == 1:
+        courses_z = 0
+    else:
+        courses_z = -1
+
+    # Calculates z-score of top instructors result
+    if len(instructors['results']) > 1:
```
alex-shen1, line 53: This logic is essentially identical to the section above; I'd refactor into something like `compute_max_zscore()` and then just pass in the course/instructor score lists as a parameter.

tcf_website/views/search.py
```diff
@@ -17,14 +18,61 @@ def search(request):
...
+        if instructors_stddev == 0:
+            instructors_stddev = 1
+        instructors_z = instructors['results'][0].get('score') - instructors_mean
+        instructors_z /= instructors_stddev
+    elif len(instructors['results']) == 1:
+        instructors_z = 0
+    else:
+        instructors_z = -1
+
+    # Likely an abbreviation if 4 letters or less
```
alex-shen1, line 69: I remember we talked about this, but can you elaborate on why we need to hardcode it this way (and expand the comment with the explanation)? Something with z-scores being weird if the returned sample size is too small?

tcf_website/templates/search/search.html
```diff
@@ -24,12 +24,21 @@ <h2>
           <div class="col-12">
             <div class="btn-group btn-group-sm" role="toolbar">
               <a class="btn bg-tcf-indigo text-white" disabled>See results for</a>
+              {% if courses_first %}
               <a href="#courses" id="course-tab" class="btn btn-light collapsed" autofocus
```
alex-shen1, line 28: My understanding is that the only difference here is which of the HTML tags has the `autofocus` attribute. If that's true, then you can simplify this logic to avoid duplicating the components themselves, like this:

```suggestion
              <a href="#courses" id="course-tab" class="btn btn-light collapsed" {% if courses_first %} autofocus {% endif %}
```

Note that if you want to see if `courses_first` is false, you'd do `{% if not courses_first %} {% endif %}`

tcf_website/templates/search/search.html
```diff
@@ -41,8 +50,13 @@ <h2>
         <div id="results" class="accordion">
 
           <!-- Course results -->
+          {% if courses_first %}
           <div id="courses" class="row mb-5 collapse show"
```
alex-shen1, line 54: Similar thing - instead of wrapping the entire block in a conditional, you can just wrap the `show` attribute.

tcf_website/templates/search/search.html
```diff
@@ -105,8 +119,14 @@ <h5><a href="{% url 'department' dept_id=dept_data.dept_id %}">{{ dept }} / {{ d
           </div>
 
           <!-- Instructor results -->
+          {% if courses_first %}
           <div id="instructors" class="row mb-5 collapse"
                aria-labelledby="instructor-tab" data-parent="#results">
+          {% else %}
+          <div id="instructors" class="row mb-5 collapse show"
```
alex-shen1, line 126: Same thing here

tcf_website/templates/search/search.html
```diff
@@ -24,12 +24,21 @@ <h2>
...
                  role="button" data-toggle="collapse" aria-expanded="true" aria-controls="courses">
                  Courses</a>
               <a href="#instructors" id="instructor-tab" class="btn btn-light collapsed"
                  role="button" data-toggle="collapse" aria-expanded="false" aria-controls="instructors">
                  Instructors</a>
+              {% else %}
+              <a href="#courses" id="course-tab" class="btn btn-light collapsed"
+                 role="button" data-toggle="collapse" aria-expanded="false" aria-controls="courses">
+                 Courses</a>
+              <a href="#instructors" id="instructor-tab" class="btn btn-light collapsed" autofocus
```
alex-shen1, line 38: Same thing here

alex-shen1 (changes requested):
Good job! I suggested a few refactors you can do for code cleanliness, but works great.

tcf_website/views/search.py
ajnye: Yep, it is sorted from top to bottom. I printed out scores when testing to find that out.

tcf_website/views/search.py
ajnye: Z-scores don't work at all if the sample size is 1 or less, because there's no standard deviation to calculate.

alex-shen1 (approved):

PR #598, opened 2022-09-13
Fix pylint

alex-shen1:
Pylint is broken on all new builds because `PYTHONPATH` isn't set. Not sure why it isn't being set when it's worked for 2 years, but I'm setting it manually. Also needed to fix a few things to make Pylint compliance back to 10/10.


jackliu612:
Thanks for the changes and fixes!

jackliu612 (approved):

Issue #236, opened 2020-11-16
Middle Eastern Studies and South Asian Studies


alex-shen1:
They've since merged, but we have empty department categories.

Issue #220, opened 2020-11-14
Data weirdness


alex-shen1:
Kelsey Johnson (astro prof) teaching ECON 2010, Mark Thomas (econ/history prof) teaching BIOL lab in the data.

Not sure why this happen

PR #624, opened 2022-11-17
Semester data overhaul

### LINKED ISSUE #403: Unify Instructor models with duplicate instances due to broken encoding
## Describe the bug
Report from @joek13:
image

## To Reproduce
You can just see the broken data on prod at any point.

## Expected behavior
We want only one instance of the Instructor model for each real-life instructor, so we can probably write a Django management script or something to unify the data. We might also want to update `load_semester` so we don't need to run the bugfix every time, or just call the script at the end.

## Screenshots
One example of the issue:

image

## Your Environment:
N/A

## Additional context
Would be good to automate the collection of professors with broken names (maybe filter the list for non-ascii characters?)

### LINKED ISSUE #461: Investigate missing instructors.
## Issue
While working on #434, I noticed that our database is missing ~800 instructors that are present in the FOIA'd UVA grades data. I think it's likely that most of these are one-time instructors or TAs leading discussion sections, so I'm going to focus on getting that PR's features finished first.

## To Reproduce
I added a flag `--log-missing-instructors` to the `load_grades`Django command, so you can run `docker exec tcf_django python manage.py load_grades ALL_DANGEROUS --log-missing-instructors`. This will create the file `missing-instructors.txt` wherever you run the command, containing all the (unique) first/last name pairs where the script tried to find an `Instructor` in our DB and didn't find any.

### LINKED ISSUE #529: load_semester command should remove any sections that don't exist in the updated version of the data
Currently, we have a script that scrapes Lou's List and adds everything listed to our database. The problem is that when a course gets removed, our data ingest logic isn't capable of removing courses that no longer exist.

For example: CS 4710 (Artificial Intelligence) was once listed as being offered in Spring 2022, but the courses got shuffled and now it is no longer being offered. However, the [CS department page](https://thecourseforum.com/department/31) on the site still lists it as being offered because we don't have logic to remove data that isn't in our updated CSV.

The easiest solution would simply just to update `load_semester.py` to delete all instances of `Section` at the start of the ingest for a file and then let the process create them again, but I'm not sure off the top of my head if that would mess with any foreign keys in our database. We'll have to do some investigating on that front.

alex-shen1:
## Status

- In Progress, and will be in progress for a while lol

## GitHub Issues addressed

- This PR closes #529, closes #461, closes #403

Seems like there's a discrepancy in the subjects available thru SIS and the ones we have.  There's a lot more subject _name_ discrepancies, but class code is probably what we actually care about (e.g. RELB is "Religion-Buddhism" in Lous/tCF but just Buddhism in SIS). This might be because I'm only doing undergrad classes right now, but it warrants more investigation.
image

## Notes
- It seems like Lou's List was manually aggregating the subjects into departments, because the SIS API doesn't provide them. Accordingly, we might have to manually do this for other ones.
- Scraped *every single undergrad class* and it took this long: 
image


alex-shen1:
This will close #220 as well because it overhauls how we do semester data collection

alex-shen1:
Might also resolve #236

alex-shen1:
May also solve #467

PR #627, opened 2022-11-25
feat(#103): toggle sidebar with mobile devices

### LINKED ISSUE #103: Mobile sidebar UX improvement
Add the functionality where the user can swipeout/away for sidebar & collapse on outside tap

barrett-ruth:
## Status

- Done (needs testing)

## GitHub Issues addressed

- This PR closes #103

## What I did

- Add JQuery to toggle mobile sidebar when anything but the sidebar and menu toggler is clicked

## Screenshots

- N/A

## Tests

- Need help here -- don't know how to actually test mobile-only features

- 320px used for mobile device width

## Merging the PR

- Who should merge this PR?
  - [] I will merge it myself
  - [x] preserve the history
  - [] squash-merge


tcf_website/templates/base/base.html
```diff
@@ -56,6 +56,19 @@
     $("#wrapper").toggleClass("toggled");
   });
 
+  /*
+    Toggle sidebar on tap with mobile devices
+  */
+  if ($(window).width() <= 320) {
```
alex-shen1, line 62: This width is too narrow to support some mobile devices

alex-shen1 (changes requested):
Functionality works great! ....When the screen is small enough.

You can test mobile features in Chrome (and probably any modern browser) by opening the developer tools with F12, then adjusting the screen size: 
image

Clicking the circled button will shift your phone to a mobile UI, and you can select specific device dimensions to test what the app will look like on those devices. For example, you can see that a Galaxy S8+ is 360 across, which is wider than the hardcoded preset of 320 that you've set, so even though the site shifts into mobile mode and displays a sidebar, your toggling functionality doesn't work.  I'm not sure how the app knows to even display the sidebar toggle in the first place, but I'd look for where that logic is and replicate it to make sure that any screen that shows a sidebar also has this toggle.

Also, just a tip for local testing: it seems that even though you can update the window dimensions dynamically, JS doesn't update what the value of window.width() is, so you need to refresh the page each time you change it.

tcf_website/templates/base/base.html
barrett-ruth: I attempted to find a screen width that would work for many devices. I'm not sure what the "perfect number" here is, could you offer some help? 480 perhaps?

alex-shen1: I'm too busy to properly investigate today, but I did some quick searching and found this in `static/base/base.css`: 
image

So it seems like 576px might be it? I'm not really sure where these numbers came from though, or even if the units used in the code are actually px, so try looking into other parts to figure out how the sidebar is configured.

barrett-ruth:
Works with Samsung Galaxy S8+ and generally <= 576 px in width.

barrett-ruth:
@alex-shen1 fixed this a while back, forgot to ping you. Should be good to merge.

alex-shen1 (approved):

Issue #530, opened 2021-10-31
Delete Account


n4mahajan:
**Is your feature request related to a problem? Please describe.**
Currently, we have no option for a user to delete their account and everything associated with it. We've had a few users in the past who wish to delete their account, so it's worthwhile to implement such a feature.

**Describe the solution you'd like**
Most likely a button or section on a user's profile page that simply allows the user to delete their account. If the user does try deleting their account, they should be prompted with a confirmation beforehand to make sure they actually delete their account. Once an account is deleted, any data associated with that account (reviews, votes, etc.) should be removed along with it. Since only UVA emails can be used for tCF, deleting an account is essentially just deleting any data associated with an account, as a user could hypothetically delete their account and then use their UVA email again to log in to tCF.

**Describe alternatives you've considered**
N/A

**Additional context**
This feature has been listed as a good first issue for new devs, but realistically a new dev will need help with this which is totally expected. It's not too overly difficult and still gives good experience with getting used to the work flow, but feel free to reach out especially since this is a relatively difficult first task to undertake.

PR #628, opened 2022-12-04
Delete account

t0mmylam:
## Status

- Done

## GitHub Issues addressed

- This PR closes
- #530 

## What I did

- Created a button on the profile page that allows a user to delete their account
- #626 was merged into this branch

## Screenshots

- Before
Screen Shot 2022-12-04 at 2 29 33 PM

- After
Screen Shot 2022-12-04 at 2 29 07 PM

## Tests

- Write a review and vote on reviews then delete the account

## Merging the PR

- Who should merge this PR?
  - [] I will merge it myself
  - [X] The last reviewer to approve can merge it
  - [] preserve the history
  - [X] squash-merge


tcf_website/templates/profile/delete_confirm_modal.html
```diff
@@ -0,0 +1,25 @@
+<!-- Delete Confirmation Modal -->
+<div class="modal fade" id="deleteProfileModal{{user.id}}" tabindex="-1" role="dialog" aria-labelledby="deleteModalLabel" aria-hidden="true">
```
alex-shen1, line 2: Just wondering -does it particularly matter if the ID of this element has the user ID instead of just being called `deleteProfileModal`? I see it's used as the data target for the modal in `profile.html`, but I can't see anything that would suggest it needs to be unique per user.

tcf_website/views/profile.py
```diff
@@ -71,3 +77,23 @@ def reviews(request):
...
+        # For security: Make sure target review belongs to the current user
+        if obj != self.request.user:
+            raise PermissionDenied(
+                "You are not allowed to delete this account!")
+        return obj
+
+    def get_success_message(self, cleaned_data) -> str:
+        """Overrides SuccessMessageMixin's get_success_message method."""
+
+        # return success message
```
alex-shen1, line 98: Not sure this comment is necessary tbh

alex-shen1 (commented):
Tested locally and it looks great! Just have a few minor comments, but the main blocker here is the pylint stuff.

tcf_website/templates/profile/delete_confirm_modal.html
t0mmylam: good point seems redundant i'll make that change

tcf_website/views/profile.py
t0mmylam: agreed

PR #629, opened 2022-12-06
Email login

jackliu612:


alex-shen1 (commented):
- Attempting to log in with an email that isn't registered will send you to `/register` without the email prefilled. In any case, it probably shouldn't send you to `/register` anyways, it should say the account doesn't exist.
  - Actually, attempting to log in with a valid `@virginia.edu` will send you directly to the email confirmation notice for some reason? But it'll give you this message
image
  - Seems like just not redirecting to the create page should work... it'd be better if we did error messages in the modal itself anyways, but that might be more difficult.
-  An improper email address will hit you with the generic "There was an error logging you in. Please make sure you're using an @virginia.edu email address." which isn't _technically_ wrong, but I feel like it should be different, like "there was an error creating your account"?
- Refreshing the email confirmation page should redirect, but just says None
image
- Attempting to create an account that already exists will say "Check your email and password" instead of an appropriate error message

tcf_core/auth_pipeline.py
```diff
@@ -65,3 +95,33 @@ def create_user(strategy, details, backend, user=None, *args, **kwargs):
         'is_new': True,
         'user': strategy.create_user(**fields)
     }
+
+def check_user_password(strategy, backend, user, is_new=False, password="", *args, **kwargs):
+    """
+    Saves passwort to user object if a new user (registering).
```
alex-shen1, line 101: password

PR #645, opened 2023-01-30
added instructor hidden in model.py and browse.py course_view

Caesarsyning:
## GitHub Issues addressed

- Removing professors who no longer want to be viewed in the tCF

## What I did

- I added the  "hidden" a boolean field with a default value False in the instructor model.py
- Added "hidden = False" in browse.py course_view function for instructor filter

## Screenshots

- Before
Screen Shot 2023-01-30 at 2 00 28 PM

Screen Shot 2023-01-30 at 1 59 06 PM

- After
Screen Shot 2023-01-30 at 2 00 19 PM
Screen Shot 2023-01-30 at 2 10 27 PM

## Testing

- Set field true for the instructor that need to be hidden in the website admin


jackliu612 (approved):
Thanks for the patch!

PR #647, opened 2023-02-07
Correct how grades are calculated

alex-shen1:
After some investigation, the current way average GPA is calculated on dev (where DFW is dropped entirely) is unacceptably skewed.

For reference, here's CS 2150, a class with notoriously high rates of DFW:
image

On dev, the new formula has caused average GPA to be massively overreported:

image

With the new formula, it's much closer:
image

Note: not sure why the CSV files changed. Had to fix the column names for some files, but I guess in the process the other data got saved over..?


tcf_website/tests/test_commands.py
```diff
@@ -71,7 +71,56 @@ def test_matching_data(self):
             'c_plus',
             'c',
             'c_minus',
-            'dfw']:
+                'dfw']:
```
alex-shen1, line 74: This space is autogenerated by autopep8. Nothing I can do about it :/

jackliu612 (approved):
Thanks for addressing my comments, looks good to me!

PR #669, opened 2023-08-28
fix(#665): Implement data backups

### LINKED ISSUE #665: Implement data backups
**Is your feature request related to a problem? Please describe.**
To avoid future issues with data loss, it would be a good idea to create some sort of automated data backup script that will save our production DB on DigitalOcean. This will ensure that user data is properly stored and accessible regardless of any issues that might arise.

It would be best to create an automated script that will run periodically and backup our production data. Ideally, the script would run multiple times a semester (and definitely once at the end) to ensure we have multiple versions of the database at various points.

**Describe alternatives you've considered**
The backups probably should not be done within DigitalOcean itself to ensure that they aren't impacted by issues with have with the service. Also, if the automation part is difficult to figure out, then just the creation of a backup script should be fine for now.

**Additional context**
N/A

barrett-ruth:
## GitHub Issues addressed

- This PR closes #665

Adds backup script to dump db file based on .env (currently doesn't do much)

- Attempting to integrate [gdrive](https://github.com/glotlabs/gdrive) with the script but having to jump through several hoops to get the necessary scopes added to GCP credentials. WIP


barrett-ruth:
Note: waiting on google to resolve authenticator credentials

PR #678, opened 2023-09-24
fix(search)!: trigram search algorithm

barrett-ruth:
This PR is a drop-in replacement of Elasticsearch in favor of PostgreSQL trigram algorithm.

@ajnye @Caesarsyning please review.

Removed all elasticsearch and related info from the DB.

***FURTHER TESTING REQUIRED - STILL SLOW ON SOME OF MY QUERIES***
AJ will also be stress testing the search code before implemented.

***NOTE***: postgres extension MUST be enabled in all environments, INCLUDING PRODUCTION:


Caesarsyning:
Two optimizations
- Set  MNEMONIC_WEIGHT to 1 when no number provided
- Only search courses when number provided

barrett-ruth:
@Caesarsyning I think we still want to search courses if no number provided. Also, the algorithm is still even slower than tCF in some examples.

barrett-ruth:
The largest search time is being bottlenecked by the `fetch_instructors` when the results have a low similarity. For example, searching obscure terms like `xyz` or `ggg`

barrett-ruth:
**NOTE**: will also have to enable extension in production

barrett-ruth:
@artiehumphreys thanks for those improvements. added my reviews above

Issue #680, opened 2023-09-25
tCF logo a 4000x4000 png image


wkaisertexas:
In production, a 4000x4000 png image is being served for the logo. 

This could easily be an SVG

PR #682, opened 2023-09-25
accessibility + icon format fixes

wkaisertexas:
## GitHub Issues addressed

#680 is closed by replacing the large png logo with an svg logo

## What I did

- Created an `new_logo.svg`
- Find and replace all `new_logo.png` with `new_logo.svg`
- Deleted `new_logo.png`

## Screenshots

Literally nothing changes. The new SVG is a pixel-perfect copy I made in Figma.

Fixing the logo size saves ~9 seconds on the lighthouse report when ran on the [thecourseforum.com/browse](https://thecourseforum.com/browser) page.

Screenshot 2023-09-25 at 1 30 27 AM

---

Apart from that, I did some **class-based refactoring** in the `index.py` views class which, in addition to being more consistent, avoid disk operations (particularly when loading the index page which could theoretically happen frequently)

Finally, a bit of accessibility with adding `aria-labels` are per the lighthouse report.


barrett-ruth:
Remove any and all changes to the django code besides the `aria-label` attributes and updating to use the svg.

There's a lot of *not good* code and styling in the codebase (to put it lightly). We don't want to accomplish two tasks in one PR.

Separate out the changes into another PR - we still definitely want them!

wkaisertexas:
Are you telling me I should actually learn how to use git and github the proper way? 

Blasphemy.

---

I will separate out the changes. Honestly, there is a lot more work to be done with accessibility (aria-labels, contrast ratios, not applying opacity when it does not need to be applied) that I will wait to group changes in a more logical way.

PR #696, opened 2023-10-15
add about and img for Rui - dev team

RuiyuZhang69:
## Screenshots

- After
IMG_1166

## Questions/Discussions/Notes

-

PR #713, opened 2023-10-20
feat(#548): about page cleanups

### LINKED ISSUE #548: Add D&M people to the site w/ LinkedIns
Simply adding D&M people to the site is pretty much the same as adding devs, but this issue also includes adding links to their LinkedIns - devs get to link their GitHubs, so for D&M people LinkedIns are roughly the closest thing. See here:
image

The idea is that where that GitHub logo is right now (it's also an `<a>` tag), there should be a LinkedIn logo that also acts as a link.

barrett-ruth:
**I totally didn't see #696.** My mistake!!! I think this UI is preferable though.

Resolves #548. 

Removes Tommy, he transferred to Columbia.

De-duplicates members in exec not in D&M.

Linkedin & GitHub icons (optionally) for all members.

Centers all members.

Before:
image

After:
image


Ann924 (changes requested):
Some members are missing a GitHub link in this cleanup even though they had one on the original site (i.e. Check Oliver).

I understand that we might want to de-duplicate the exec members, but I feel like it would be best to be consistent in how we handle that for the engineering and design/marketing teams. For that reason, I personally lean towards keeping the duplicates. 

I think the engineering team members still need to be cleaned out a bit more. If Raffi, Samyak, etc. aren't returning this semester they should be moved to "Past Contributors." Similarly, if Tommy made any contributions, he should also be in "Past Contributors". We might want to hold off on merging this PR until a couple of meetings later just to be clear on who is returning.

barrett-ruth:
AJ was fine with the de-dups, we should discuss next meeting.

Moved tommy, raffi, samyak to alums. fixed oliver's github

barrett-ruth:
@Ann924 now that photos are resized, what do you want to do with this? I think we should just make a decision for the duplicates and stick to it. I'll update the pfps accordingly.

Currently, only the teach team is de-duplicated from exec. Otherwise every else is duplicated for their respective department.

PR #716, opened 2023-10-21
remove extra quote in review-redirect

Ann924:
Removed extra quote in tcf_website\templates\course\course_professor.html for review redirect.

(Originally missed when I approved and merged into dev, caught by Barrett).


barrett-ruth (approved):
noice

PR #717, opened 2023-10-21
Add About & PFP for Jay Lalwani

Jay-Lalwani:
## What I did

-Added About & PFP for Jay Lalwani

## Screenshots

- After
image


Ann924 (approved):
Looks great!

PR #718, opened 2023-10-22
add logout call before deleting user profile

kmd2zjw:
## GitHub Issues addressed

- This PR closes

## What I did

- I added a logout from the Auth library. My thinking is that this will flush out all the data when a profile is deleted. 
- I also think the initial issue was poorly described, I had encountered some initial confusion since deleting the profile was logging users out. So I just added the logout call in order to address the issue

## Screenshots

- Before
- After

## Testing

-


barrett-ruth (approved):
reviewed in person

Issue #700, opened 2023-10-15
Cursor on FAQ Dropdown


Ann924:
## Describe the bug
When the cursor hovers over the dropdown items in FAQ, it does not appear as a pointer.

## To Reproduce
Steps to reproduce the behavior. For example:
1. Go to https://thecourseforum.com/
2. Scroll down to the FAQ section
3. Hover mouse over any of the dropdown items
4. See cursor error

## Expected behavior
The cursor should be a pointer cursor rather than a text cursor when the mouse hovers over any of the elements.

## Screenshots
Note the form that the cursor takes when it is hovering over the text.
![image](https://github.com/thecourseforum/theCourseForum2/assets/78228879/e2a18d35-aa8a-45ee-8982-64512569ae7c)

## Your Environment:
 - OS: Windows 10
 - Browser Chrome

PR #719, opened 2023-10-22
Make cursor pointer on FAQ dropdown

junholee6a:
## GitHub Issues addressed

- This PR closes https://github.com/thecourseforum/theCourseForum2/issues/700

## What I did

- Added a CSS statement to change the cursor to a pointer

## Screenshots

- Before
Screenshot 2023-10-22 at 3 41 21 PM

- After
Screenshot 2023-10-22 at 3 42 10 PM

## Testing

1. Go to https://thecourseforum.com/
2. Scroll down to the FAQ section
3. Hover mouse over any of the dropdown items
4. See cursor change to a pointer

- The cursor will turn into a pointer even when it's to the right of the FAQ text rather than directly on the text (this is the case in the "after" screenshot as well). I think this is valid because the FAQ question is still clickable when the pointer is to the right of the question.


barrett-ruth (approved):
lgtm

PR #704, opened 2023-10-17
Browse view cleanup

wkaisertexas:
## GitHub Issues addressed

This fixes the issue of redundant templates for `school.html` and `grouped_schools.html`

## What I did

- Combined Templates
- Fixed the visual glitch where closing is not smooth

PR #728, opened 2023-10-27
Resize alumni photos

vsong15:
## GitHub Issues addressed

- This PR closes

## What I did
Resized alum photos not including founders to be 800x800.
-

## Screenshots

- Before
image
- After
image

## Testing

-


barrett-ruth:
Thanks for this whole crap-loading of responding. Don't mean to make you go through all of this again but some photos are still too big (or >800x800 - not sure if that's ok or not). For example, file Chai's photo is 2160x2160:

image

Can you go back through and check all these photos? I'd definitely recommend using a script to automate this after all the work you've done like below (this only works on Mac/Linux btw):

```bash
for file in tcf_website/static/about/team-pfps/*; do
    echo "$file has file info: <$(file "$file")>";
done
```

vsong15:
Hi Barrett, thanks for the feedback, I will go back and fix those alum photos. I think you uploaded a different photo instead of Chai's photo for the example.

vsong15:
I finished resizing the alum-pfps and team pfps. They should all be 800x800 now. Should the founders also be resized to be square or is it fine. I think that if it is resized to square it might have to be larger than 800x800 since it takes up more space, but it also looks fine as is.

wkaisertexas:
```console
for image in *.$1; do
  base=${image%.*}
  cwebp -resize 256 256 $image -o $base.webp
  # cwebp -resize 256 -1 $image -o $base.webp
done
```
I did not know this was on the docket, but here was the script which I used to convert images. There are a bunch of different file types present in the codebase so this allows you to say `bash myscript.sh jpeg` and `bash myscript.sh png` to convert the respective images.

This uses the `cwebp` which is [Google's Tool to convert WebP images](https://developers.google.com/speed/webp/docs/cwebp) which I just installed through [Homebrew](https://brew.sh/) which is just a package manager for MacOs.

barrett-ruth:
Left your branch, right tcf:

image

The spacing/CSS could definitely be improved. tCF itself doesn't look perfect as it, but make sure to match the current appearance at least (I think the right looks better, let me know if you think otherwise).

wkaisertexas:
@barrett-ruth 

Your branch on the right looks weird. I updated the css for this page to include a padding of `1.25 rem`. Could your browser still have the old CSS cached which could lead to a glitch like that? 

Screenshot 2023-11-02 at 11 09 24 PM

Also, I think this is the wrong thread. The actual pr of interest is #704

vsong15:
I finished creating the python script to automate resizing the images.

ldkohler:
The only issue I see now is with some of the images being square and some being rectangular, which just causes so weird visual inconsistencies especially with the text below the images like here: 
image

idk if there much you can do with this, but just maybe try scaling up some of the square ones and cropping the sides so that they are all rectangular and consistently framed. i would just play around with that to see if there anything that works and it looks odd then we can just go with how it currently is

ldkohler:
looks good, white bars on the old images are a little weird but thats not really anything we can do about that really

ldkohler:
looks good, white bars on old images are a little weird but we cant really do much about that

barrett-ruth (approved):
lgtm

PR #741, opened 2023-10-31
January/Spring 2024 course update

Ann924:
Added CSVs from Louslist with course offering data for January 2024 and Spring 2024 and updated database.


barrett-ruth (approved):
lgtm

PR #743, opened 2023-11-04
perf(docker): cache installation layers

barrett-ruth:
Separation out installation steps in docker to improve caching and performance.

General cleanups as well.


wkaisertexas (approved):
LGTM

PR #745, opened 2023-11-06
more modest page sizes

wkaisertexas:
## GitHub Issues addressed

- @ajnye conversation about query optimizations 

## What I did

Adjusted API to not have an unlimited page size. Makes it so tCF can't get taken down by like four requests 

This comes after looking at the digital ocean where queries for related course information would take ~1 second on average and up to 30 seconds max. Shocking considering most queries could operate at <1ms.

### DIFF
--- a/tcf_website/api/paginations.py
+++ b/tcf_website/api/paginations.py
@@ -4,6 +4,6 @@ from rest_framework.pagination import PageNumberPagination
 
 class FlexiblePagination(PageNumberPagination):
     """Just a custom pagination class"""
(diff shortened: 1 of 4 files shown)


wkaisertexas:
Screenshot 2023-11-05 at 10 58 06 PM
Screenshot 2023-11-05 at 10 58 40 PM

These were the two queries which looked the worst. Still not great but better than nothing. 

> Forgive my terrible screenshots

wkaisertexas:
@barrett-ruth tf?

PR #756, opened 2023-11-12
basic cache setup

wkaisertexas:
## What I did

- Added a basic memcache and dummy cache to settings (https://docs.djangoproject.com/en/4.2/topics/cache/#memcached)
- Allows the use of @cache_page decorators, template fragment decorators, and fine-grained cache control

## Discussion

We have [Cachalot](https://django-cachalot.readthedocs.io/en/latest/) which caches at the request level. However, the ability to use fine-grained cache control could be very useful as we start rolling out more advanced features.

The ability to use cache provides a useful tool in the toolboxes of tCF developers.


wkaisertexas:
#830 added a memcache which worked pretty well

PR #759, opened 2023-11-15
feat(format):

barrett-ruth:
TODO: use old strategy, make more obvious, and integrate current pre-commit config into it.
Add `pre-commit` config for all developers and format the codebase accordingly.

Tools used: black, isort, prettier

See `.pre-commit-config.yaml` for details.

TODO: add README w/ details
TODO: developer announcement


barrett-ruth:
@wkaisertexas this PR offers a good transition to using the tool `pre-commit`, but it runs into versioning issues as you can see in the CI. Let me know your thoughts, but I was struggling trying to install outdated packages with versions of dependencies more than 4 years old.

This I think is its own problem, but would FIRST require updating the versioning of the entire codebase, THEN merging this.

PR #763, opened 2023-11-21
Microsoft login

ldkohler:
## GitHub Issues addressed

- Microsoft auth added
- Sign in UI fixes

## What I did

- Added Microsoft authentication to sign in and new sign in button to modal
- Partially redid the UI for the sign in but we can comeback to it later
- New auth helper file for microsoft auth
- Removed create an account button since it's obsolete now

## Screenshots

- Before
image

- After
image

- We need to make sure we get ITS approval for the microsoft auth to actually work for all users
- Microsoft auth currently brings us to this:
image


ajnye:
Will not be possible unless given permission from UVA ITS.

PR #764, opened 2023-11-23
server 500 fix

RuiZhang-kwf8:
## GitHub Issues addressed
-Server 500 error upon getting outlook email. 

## What I did

- Changed Social Auth Pipeline to add email validation/send function since its essentially the same thing.

## Screenshots

- Before

- After
in base.py added     'tcf_core.auth_pipeline.validate_email': 

Screenshot 2023-11-22 at 9 10 53 PM

## Testing
Testing cant be done because it works locally so we probably need to deploy to see if anything fixes. 

-

### DIFF
--- a/tcf_core/settings/base.py
+++ b/tcf_core/settings/base.py
@@ -166,6 +166,7 @@ SOCIAL_AUTH_PIPELINE = (
     'social_core.pipeline.social_auth.social_details',
     'social_core.pipeline.social_auth.social_uid',
     'tcf_core.auth_pipeline.auth_allowed',
+    'tcf_core.auth_pipeline.validate_email',
     'social_core.pipeline.social_auth.social_user',
     'social_core.pipeline.user.get_username',
     'social_core.pipeline.mail.mail_validation',


tcf_core/settings/base.py
```diff
@@ -166,6 +166,7 @@
     'social_core.pipeline.social_auth.social_details',
     'social_core.pipeline.social_auth.social_uid',
     'tcf_core.auth_pipeline.auth_allowed',
+    'tcf_core.auth_pipeline.validate_email',
```
junholee6a, line 169: image


I think the issue is that `tcf_core.auth_pipeline.validate_email` isn't meant to be used in `SOCIAL_AUTH_PIPELINE`. According to [docs](https://python-social-auth.readthedocs.io/en/latest/pipeline.html#email-validation), to set up email validation you:
1) Add `social_core.pipeline.mail.mail_validation` to `SOCIAL_AUTH_PIPELINE` (already being done [on this line](https://github.com/thecourseforum/theCourseForum2/blob/52cb9f74a8c9649b44e4dd0bb2bd752b3618dae1/tcf_core/settings/base.py#L172))

2) Add a custom function to `SOCIAL_AUTH_EMAIL_VALIDATION_FUNCTION`. In the existing code, `tcf_core.auth_pipeline.validate_email` is used for this [on this line](https://github.com/thecourseforum/theCourseForum2/blob/52cb9f74a8c9649b44e4dd0bb2bd752b3618dae1/tcf_core/settings/base.py#L157)

ajnye:
Does not work, need a more in-depth fix.

PR #775, opened 2023-12-11
feat(tooling): add optional nvm, pyenv versioning

barrett-ruth:
add versioning files for nvm and pyenv

does not impact production, just if you use these tools


barrett-ruth:
to be updated in incoming versioning

Caesarsyning (approved):

PR #784, opened 2024-01-27
fix(docker): make wait-for-it compatible w/ prod db access

barrett-ruth:
allow access to prod db using `.env.prod` details.

new "ENVIRONMENT" field located in all .env files (a default ENVIRONMENT has no impact, ignoring "wait-for-it".


ajnye (approved):
Looks good!

PR #796, opened 2024-02-08
Search autocomplete system

ajnye:
## GitHub Issues addressed

- This PR closes

Adds an autocomplete feature to the search bar for classes.

## Screenshots

- Before
- After

## Testing

-


barrett-ruth:
The main search dropdown is not the same width as the search box:
image

And is also a different color (a white instead of a gray) than the main menu:
image

@artiehumphreys i think the search bar should extend up until the search icon, but not including.

As for the dropdown, I think it should be white.

Let's add these CSS changes then consult design for last tweaks/how it could be improved.

PR #804, opened 2024-02-16
fix(gae): update deployment action from deprecated versions to newest versions incorporating Node.js version 20

Jay-Lalwani:
image

Addresses this GAE warning


barrett-ruth (approved):
w ty

Issue #802, opened 2024-02-14
Email Verification Server 500 Error


josephAttia:
## Describe the bug

When verifying my email I am redirected to a Server 500 error page that cannot verify my email.

URL in question: https://thecourseforum.com/oauth/complete/email/?verification_code=######&partial_token=#######

## To Reproduce

Steps to reproduce the behavior

1. Signup with a new email
2. Click on the verification email link
3. Redirected to 

## Expected behavior

Email gets verified and I am redirected to my main account

## Screenshots

![image](https://github.com/thecourseforum/theCourseForum2/assets/61055258/719c5614-07ed-458d-b50b-85400542a32b)

## Your Environment:

- OS: Windows
- Browser Chrome

## Additional context
N/A

PR #807, opened 2024-02-19
Fix email verification

junholee6a:
Problem: Whenever someone tried to create an account by email and password (not sign in with Google), clicking on the email verification link would cause a Server 500 Error, and the account would not be created.

The cause of the problem: Between the time the user registers a new account and the time the user clicks on the email verification link, the user's data is stored in the database table `social_auth_partial`. For an unknown reason, the production server behaves as if the verification link was visited BEFORE the user clicks the link. This removes the user's data from the database, such that their account can't be created when they click the link.

Solution: Make the user's data persist in the database until the user manually enters their graduation year, then delete the entry from `social_auth_partial`.

## GitHub Issues addressed

- This PR closes https://github.com/thecourseforum/theCourseForum2/issues/802

## Testing

- Successfully created an account locally with email login, and verified that none of the new partial tokens remain on the database after account creation.
- Successfully created an account on the Heroku dev server with email login.
- To test:
    - If you already have an account, delete your account before recreating it by My Profile > Delete Profile.
    - To check that the user's data is properly removed from the database after account creation, `psql` into the database and run the query `SELECT * FROM social_auth_partial;`

### DIFF
--- a/tcf_core/auth_pipeline.py
+++ b/tcf_core/auth_pipeline.py
@@ -4,6 +4,7 @@
 # pylint: disable=line-too-long
 """Custom authentication pipeline steps."""
 
+from functools import wraps
 from django.conf import settings
 from django.contrib.auth.password_validation import validate_password
 from django.core.exceptions import ValidationError
@@ -11,6 +12,7 @@ from django.core.mail import send_mail
 from django.shortcuts import redirect, render
 from django.urls import reverse
 from social_core.pipeline.partial import partial
+from social_core.strategy import BaseStrategy
 
 from tcf_website.models import User
 
@@ -44,12 +46,33 @@ def password_validation(backend, details, request, response, *args, **kwargs):
             return redirect("/login/password_error", error=True)
     return {"password": response.get("password")}
 
+# Make BaseStrategy.validate_email() always return true, so that
+# social_core.pipeline.mail.mail_validation can be run multiple times
+def always_true(*args, **kwargs):
+    return True
+BaseStrategy.validate_email = always_true
 
-@partial
+# Wrapper over strategy.clean_partial_pipeline() to implement partial token persistence
+def partial_token_persistence(orig_func):
+    @wraps(orig_func)
+    def new_func(self, *args, **kwargs):
+        if (self.session.get('persist_partial_token', default=False)):
+            # Persist partial token
+            pass
+        else:
+            # Delete partial token
+            orig_func(self, *args, **kwargs)
+    return new_func
+BaseStrategy.clean_partial_pipeline = partial_token_persistence(BaseStrategy.clean_partial_pipeline)
+
 def collect_extra_info(
     strategy, backend, request, details, user=None, *args, **kwargs
 ):
     """Collect extra information on sign up."""
+
+    # Disable partial token persistence
+    strategy.session_set('persist_partial_token', False)
+
     if user:
         return {"is_new": False}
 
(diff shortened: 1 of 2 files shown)


tcf_core/auth_pipeline.py
```diff
@@ -44,24 +45,50 @@ def password_validation(backend, details, request, response, *args, **kwargs):
             return redirect("/login/password_error", error=True)
     return {"password": response.get("password")}
 
+# Make BaseStrategy.validate_email() always return True, so that
+# social_core.pipeline.mail.mail_validation can be run multiple times
+BaseStrategy.validate_email = lambda *args, **kwargs: True
```
junholee6a, line 50: This is needed because the `validate_email` method prevents `social_core.pipeline.mail.mail_validation` from being able to run more than once.

tcf_core/auth_pipeline.py
```diff
@@ -44,24 +45,50 @@ def password_validation(backend, details, request, response, *args, **kwargs):
...
+    @wraps(orig_func)
+    def new_func(self, *args, **kwargs):
+        if self.session.get('persist_partial_token', default=False):
+            # Persist partial token
+            pass
+        else:
+            # Delete partial token
+            orig_func(self, *args, **kwargs)
+    return new_func
+BaseStrategy.clean_partial_pipeline = partial_token_persistence(BaseStrategy.clean_partial_pipeline)
```
junholee6a, line 63: This modifies the method `BaseStrategy.clean_partial_pipeline` so that we can control whether partial tokens are deleted or persisted.

tcf_core/auth_pipeline.py
```diff
@@ -44,24 +45,50 @@ def password_validation(backend, details, request, response, *args, **kwargs):
...
 
-@partial
```
junholee6a, line 48: Instead of creating a new partial token for `collect_extra_info`, we'll persist and reuse the partial token from `social_core.pipeline.mail.mail_validation`.

tcf_website/views/auth.py
```diff
@@ -72,7 +72,12 @@ def collect_extra_info(request, method):
 
             # once we have the grad_year stashed in the session, we can
             # tell the pipeline to resume by using the "complete" endpoint
-            return redirect(reverse("social:complete", args=[method]))
+            return redirect(reverse("social:complete", args=[method])
```
junholee6a, line 75: Previously, when the user set their graduation date, the auth pipeline would directly run `tcf_core.auth_pipeline.collect_extra_info`. Now that `tcf_core.auth_pipeline.collect_extra_info` uses the same partial token as `social_core.pipeline.mail.mail_validation`, when the user sets their graduation date the pipeline starts at `mail_validation`. The `verification_code` and `partial_token` must be in this redirect so that `mail_validation` finishes successfully.

tcf_website/views/auth.py
```diff
@@ -72,7 +72,12 @@ def collect_extra_info(request, method):
...
+                            + "?verification_code="
+                            + request.GET['verification_code']
+                            + "&partial_token="
+                            + request.GET['partial_token']
```
junholee6a, line 79: Previously, when the user set their graduation date, the auth pipeline would directly run `tcf_core.auth_pipeline.collect_extra_info`. Now that `tcf_core.auth_pipeline.collect_extra_info` uses the same partial token as `social_core.pipeline.mail.mail_validation`, when the user sets their graduation date the pipeline starts at `mail_validation`. The `verification_code` and `partial_token` must be in this redirect so that `mail_validation` finishes successfully.

ajnye (approved):
Looks good, thanks a ton for fixing this!

PR #810, opened 2024-02-21
Replace global monkey patching

junholee6a:
## GitHub Issues addressed

Addresses errors in Google App Engine, likely caused by monkey patching in the global scope which become undone by GAE re-importing packages.

## Testing

- Test email registration in GAE, which is the only environment we've been able to reproduce this error


ajnye (approved):
Looks good - let's give it a shot.

PR #815, opened 2024-03-12
Review Drive CSV Upload

ajnye:
## GitHub Issues addressed

- This PR closes

## What I did

- Adds a script and CSV for the Fall 2023 review drive to upload it to the database. Uses a dummy account for all emails with no tCF accounts or uses the user's account if it exists. If the review text exists, do not write to database.

## Screenshots

- Before
- After

## Testing

- Ensure the code would not crash or error partway through if done correctly, and that there's no extra checks needed. Also see if checks like review text equality should be changed.

-


tcf_website/management/commands/upload_csv_review_drive.py
```diff
@@ -0,0 +1,174 @@
...
+        year = options["year"]
+
+        # Get semester object from args
+        semester_query = Semester.objects.filter(season=season.upper(), year=year)
+        if semester_query.count() <= 0:
+            print('ERROR: Semester not found')
+            return
+        semester = semester_query.first()
+
+        # Needs to be created ahead of time in the db
```
barrett-ruth, line 66: What are we doing to ensure this user will always be created?

tcf_website/management/commands/upload_csv_review_drive.py
```diff
@@ -0,0 +1,174 @@
...
+            # Use account if found. Otherwise, use dummy user
+            account_query = User.objects.filter(email=email)
+            if account_query.count() > 0:
+                if verbose:
+                    print('Account already found for', email)
+                account = account_query.first()
+            else:
+                account = dummy_account.first()
+
+            subdepartment_query = Subdepartment.objects.filter(mnemonic__iexact=mnemonic)
```
barrett-ruth, line 100: I think stuff like this (and course_query, and account_query) should prob be inlined, and have a comment (if necessary):

```python
# account found
if User.objects.filter(email=email).count() > 0:
    ...
```

just a preference for helping these scripts be a more readable, they get super clunky

tcf_website/management/commands/upload_csv_review_drive.py
```diff
@@ -0,0 +1,174 @@
...
+from tcf_website.models import *
+from datetime import datetime
+import csv
+import os
+
+DATA_DIR = "tcf_website/management/commands/review_drive_responses/"
+
+class Command(BaseCommand):
+
+    # Run this coomand using `sudo docker exec -it tcf_django python3 manage.py upload_csv_review_drive [filename] [season] [year]`
```
barrett-ruth, line 12: Why is this here - I don't think we should comment stuff like this.

I've done quite a lot in the past but I think if a dev hasn't learn how Django commands work (and how to user docker) there's a fundamental lack of understanding somewhere else.

tcf_website/management/commands/upload_csv_review_drive.py
ajnye: The command does not run if the user doesn't exist, lines 67-70 covers this.

tcf_website/management/commands/upload_csv_review_drive.py
ajnye: I think it's fine to leave as is, many of the other commands have this in comments, I see no reason to not follow consistent formatting of commands.

tcf_website/management/commands/upload_csv_review_drive.py
barrett-ruth: Yes - my point is what if, somehow, the user is *not* created and the command is run. Do we want to create the dummy user instead of just failing?

ajnye: I was thinking of doing that, but we don't want the user password to be in the code - it's open source. We can consider adding it to the env if it's important enough, but after it gets created in the database once, there's no reason to have code to auto-add it. If you think it's important enough, I'll connect it to the .env to auto-create a user.

tcf_website/management/commands/upload_csv_review_drive.py
```diff
@@ -0,0 +1,213 @@
...
+
+        for i, line in enumerate(csv_file):
+            if i == 0:
+                continue
+
+            # Strip needed to remove random whitespace for exact match
+            email = line[1].strip()
+            mnemonic = line[2].strip()
+            num = int(line[3].strip())
+            instructor_arr = line[4].strip().split(" ", 1)
```
barrett-ruth, line 89: this has been renamed from `instructor` but is not renamed elsewhere

tcf_website/management/commands/upload_csv_review_drive.py
barrett-ruth: If you don't think it's necessary, go ahead and resolve this.

Let's say we have something irrecoverable happen to our DB - I think all of our commands and processes should be functioning as expected and not require us to reimplement changes.

For example, when adding trigram to the DB, I made sure to have a migration that enables the extension (even though, realistically, we should never need to enable the extension again).

tcf_website/management/commands/upload_csv_review_drive.py
ajnye: It is renamed, if it wasn't it would be called through instructor[0] or instructor[1]. instructor as a variable is now the object called from the query.

barrett-ruth: Seems like the `if verbose` print statement prints it before it is defined.

tcf_website/management/commands/upload_csv_review_drive.py
ajnye: Added an auto-creation from the .env.

tcf_website/management/commands/upload_csv_review_drive.py
barrett-ruth: @ajnye

barrett-ruth (approved):

ajnye:
Not finished, forgot to make the email field in reviews optional/not-nullable.

.github/workflows/continuous-integration.yml
```diff
@@ -29,6 +29,11 @@ env:
   # email for account verification
   EMAIL_HOST_USER: ${{ secrets.EMAIL_HOST_USER }}
   EMAIL_HOST_PASSWORD: ${{ secrets.EMAIL_HOST_PASSWORD }}
+  # review drive account information
```
barrett-ruth, line 32: Is this necessary to include in CI? (same for app.yaml.template, master-to-app-engine.yaml)

tcf_website/management/commands/import_csv_review_drive.py
```diff
@@ -0,0 +1,227 @@
...
+                    amount_reading + amount_writing + amount_group + amount_homework
+            )
+
+            created = datetime.strptime(line[0].strip(), "%m/%d/%Y %H:%M:%S")
+            modified = datetime.now()
+
+            if account == dummy_account:
+                review_email = email
+            else:
+                review_email = ""
```
barrett-ruth, line 201: We setting this to an empty string because the user doesn't have an email, right - I think a comment clarifying this can't hurt.

tcf_website/migrations/0012_alter_review_email.py
```diff
@@ -0,0 +1,18 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tcf_website', '0011_review_email'),
+    ]
+
+    operations = [
```
barrett-ruth, line 12: We should merge these two migrations into one.

barrett-ruth (changes requested):

tcf_website/management/commands/import_csv_review_drive.py
barrett-ruth: Also, it seems like we can avoid `review_email` all together - see comment on ~line 114

tcf_website/management/commands/import_csv_review_drive.py
```diff
@@ -0,0 +1,227 @@
...
+            instructor_arr = line[4].strip().split(" ", 1)
+
+            if verbose:
+                print(email, mnemonic, num, instructor_arr)
+
+            # Use account if found. Otherwise, use dummy user
+            account = User.objects.filter(email=email).first()
+            if account is not None and verbose:
+                print("Account already found for", email)
+            elif account is None:
```
barrett-ruth, line 114: Here, we can just set `email = ""`, then use `email` instead of `review_email`.

.github/workflows/continuous-integration.yml
ajnye: I think it would be necessary if you were to run the script from prod given that it seems to take env information from github secrets on prod instead of the .env (I don't think prod has an .env) so I think it's probably necessary.

tcf_website/migrations/0012_alter_review_email.py
ajnye: Seems like you shouldn't delete and merge migrations since they're like GitHub commits, and the only way to merge would be to delete and remake them.
https://stackoverflow.com/questions/28404461/can-i-delete-the-django-migration-files-inside-migrations-directory
I don't think it changes efficiency much to have both migrations, since they're only ever run once.

tcf_website/management/commands/import_csv_review_drive.py
ajnye: Changed it to use a ternary if statement later on instead of review_email.

barrett-ruth (approved):

PR #834, opened 2024-04-14
Password reset

junholee6a:
## GitHub Issues addressed

- Adds password reset and change features

## Screenshots

- Before
Screenshot 2024-04-14 at 1 03 06 PM

Screenshot 2024-04-14 at 1 03 22 PM

- After
Screenshot 2024-04-14 at 1 02 50 PM

Screenshot 2024-04-14 at 1 02 31 PM

## Testing

- Tested resetting a password while logged out and changing a password while logged in

- Password reset - used when logged out
- Password change - used when logged in
- I used the default Django views: https://github.com/django/django/blob/main/django/contrib/auth/views.py
- The current templates for password reset/change are kinda ugly. They can be replaced by adding `template_name` args like so in `urls.py`:
```python
path(
    "accounts/password_reset/",
    auth_views.PasswordResetView.as_view(template_name="the_new_template_name"),
    name="password_reset"
),
```


barrett-ruth (commented):
solid

barrett-ruth (approved):
lgtm

PR #840, opened 2024-04-17
Search indexing

barrett-ruth:
This PR implements a hybrid reverse indexing and trigram similarity search algorithm.

Search vectors are used for courses.
Trigram similarity is used for instructors for more fine-grained matching.

Downsides:
Cold-start indexing times for generating reverse indices

# TODO
- Weighting courses versus instructors (weights should be based on *testing*)
- Speed: TriGram searches are still run every time on instructors (is this superior to no trigram?)


artiehumphreys (dismissed):
LGTM

ajnye (approved):
Looks good, let's test it in prod

PR #861, opened 2024-06-09
Course Page Refactor

artiehumphreys:
### Course Page Refactor
**Goal: Use Django URLs to handle the sorting of instructors within the course page instead of JS**
TODO still:

- Configure sorting of instructors (alphabetical last name by default)
- Configure sorting buttons with new URLs
- Deprecate `bind-sort.js` and `bind-show-hide.js`


tcf_website/views/browse.py
```diff
@@ -91,87 +91,48 @@ def course_view_legacy(request, course_id):
...
-                        :-1
-                    ].split(",")
-        if None in i.section_nums:
-            i.section_nums.remove(None)
+    # Note: Could be simplified further
+
+    for instructor in instructors:
+        if instructor.section_times[0] and instructor.section_nums[0]:
+            instructor.times = {
+                str(num): times[:-1].split(",")
```
barrett-ruth, line 130: does this key need to be converted to a str?

tcf_website/views/browse.py
```diff
@@ -181,8 +142,8 @@ def course_view(request, mnemonic, course_number):
     # So created a dictionary on the fly to minimize database access
     semesters = {s.id: s for s in Semester.objects.all()}
     for instructor in instructors:
-        instructor.semester_last_taught = semesters.get(
-            instructor.semester_last_taught
+        instructor.semester_last_taught = str(
```
barrett-ruth, line 145: I'm not sure, but I swear this code is completely unecessary. This field should be set on the backend as we discussed

tcf_website/views/browse.py
```diff
@@ -91,87 +91,48 @@ def course_view_legacy(request, course_id):
...
+    course_number: int,
+    instructor_age: str = str(Semester.latest()),
+):
     """A new Course view that allows you to input mnemonic and number instead."""
 
     # Clears previously saved course information
-    request.session["course_code"] = None
-    request.session["course_title"] = None
-    request.session["instructor_fullname"] = None
+    request.session.flush()
```
barrett-ruth, line 103: nice

tcf_website/templates/common/rating_card.html
```diff
@@ -40,7 +40,7 @@ <h5>{{ subtitle }}</h5>
                   <i class="fas fa-chart-bar"></i>&nbsp;GPA
                 </small>
                 <p class="mb-0 info" id="gpa">
-                  {% if gpa %}
+                  {% if gpa and gpa <= 5%}
```
barrett-ruth, line 43: explicitly check for both <= 5 and >= 0 for rating,difficulty, and gpa

tcf_website/models/models.py
```diff
@@ -477,6 +489,106 @@ def review_count(self):
...
+            "gpa": "gpa",
+            "rating": "rating",
+            "difficulty": "difficulty",
+            "last_taught": "semester_last_taught",
+        }.get(sortby, "semester_last_taught")
+
+        order_prefix = "" if reverse else "-"
+
+        if recent:
+            instructors = instructors.filter(is_teaching_this_semester=True)
```
barrett-ruth, line 587: is there any way to pass this recent flag to get_instructors_and_data, so we don't end up fetching a larger queryset than necessary?

May be a stupid microoptimization - up to u

tcf_website/models/models.py
```diff
@@ -477,6 +489,106 @@ def review_count(self):
...
+        )
+        return instructors
+
+    def sort_instructors_by_key(
+        self,
+        course,
+        latest_semester,
+        recent,
+        order,
+        sortby: str = "last_taught",
```
barrett-ruth, line 570: If when grabbing the field from `sort_field` you default to "semester_last_taught", there's no point in having a default value here.

tcf_website/models/models.py
```diff
@@ -477,6 +489,106 @@ def review_count(self):
...
+                            filter=Q(review__course=course),
+                        )
+                        + Avg(
+                            "review__recommendability",
+                            filter=Q(review__course=course),
+                        )
+                    )
+                    / 3,
+                    # invalid value for if there are no reviews
+                    Value(6) if reverse else Value(0),
```
barrett-ruth, line 527: update this as we said

barrett-ruth (changes requested):

PR #874, opened 2024-09-01
Docker cache

### LINKED ISSUE #871: [Docker] Fix Dependency Hot-Reloading
{Python, JS} dependencies should only be re-installed when updates to {requirements.txt,package.json} are made.

Leverage the following approach to ensure dependencies are properly updated.

https://stackoverflow.com/questions/34398632/docker-how-to-run-pip-requirements-txt-only-if-there-was-a-change

barrett-ruth:
Resolves #871.

Caches docker/node dependencies on update.

Projects can now be built with updated dependcies via `docker compose up --build` (documentation updated).

## Example

```sh
echo asdf >> requirements.txt
docker compose up --build  # all application layers are cached but python dependencies are re-built
```


visva-murali (approved):
I tested it on my local and it works as intended and it runs. Changed the requirements.txt file and it downloaded. When I ran it again, it used its cache and did not download again.

PR #875, opened 2024-09-01
feat(about): make tino an alum

barrett-ruth:
Talked to Tino today, he said he won't be coming to meetings.


barrett-ruth:
@ajnye

PR #880, opened 2024-09-28
Padded Bar Chart Display

alex-yung-github:
- This PR closes

## What I did
Padded bar chart to account for overflowing number bug in issue 826

## Screenshots

- Before
(On the left is MATH1140 Monika and on the right is ECON 2010 Elzinga)
image
image

- After
image
image

## Testing

Tested on:
- Largest erroring classes (Econ 2010 Doyle and Elzinga)
- Other known erroring classes: (CS 2100, GDS3114
- Class with bar at exact value of max bar chart value (MATH 1140)


barrett-ruth (approved):
yup

PR #897, opened 2024-10-20
Improved Indexing for Instructors

artiehumphreys:
## Completed:

Added a composite index for efficient searching over first, last, and both first and last name fields within the instructor model. I don't believe that we need to change how we query the instructor model. 

## Testing

Some testing on a larger scale may be required.


tcf_website/models/models.py
```diff
@@ -347,6 +348,9 @@ def average_gpa(self):
             models.Avg("average")
         )["average__avg"]
 
+    class Meta:
+        indexes = [Index(fields=["first_name", "last_name"])]
```
barrett-ruth, line 352: is this composite/individual?? how does this work?

artiehumphreys: This is a composite index that allows for efficient searches over the first name and last name column (doing a logairthmic factor of reads since you can binary search through (name, primary key) pairs.)  But, upon researching, I've found that this index is efficient for first and last name searches and just first name searches, but searching for just the last name is inefficient. 

"However, a query like `SELECT * FROM users WHERE position = 'Top';` will NOT benefit from this composite index because position is the second field. A composite index sorted by class-position cannot be used to quickly find a record by position" ([Single vs Composite Indexes in Relational Databases](https://user3141592.medium.com/single-vs-composite-indexes-in-relational-databases-58d0eb045cbe)).

I think that we should experiment with trigram indices instead due to this limitation, but I'll do some more research.

barrett-ruth: You're saying we shouldn't have any indices here??

And why couldn't we index both separately and then together? would that not fix the issue?

artiehumphreys: We could do that, but the implementation is weird and fetching results won't be as simple. For a combined GIN index, we need to write the SQL ourselves since that functionality is not built into Django. When querying for results, we need to search over first names, last names, and flast names to find the best matches.

PR #908, opened 2024-10-23
Cache CI Dependencies

barrett-ruth:
- **revert: gcp on 3.11**
- **feat(ci): cache deps betwixt pylint, django**
- **fix(ci): format**

*Currently does not actually speed up CI*...


barrett-ruth:
Somehow doesn't speed up anything.

PR #910, opened 2024-10-24
fix(middleware): don't fetch DNE fields

barrett-ruth:
Fetch only existing fields from `request.META` Django request objects as it is not guarateed they exist.


junholee6a (approved):
lgtm

PR #912, opened 2024-10-25
Add Indices to Search

barrett-ruth:
TriGram/Gin Indices were never actually added to search - do that.


artiehumphreys (approved):
LGTM. Next PR needs to remove the elastic search structure on the search results.

PR #926, opened 2024-11-04
Search/fetch courses

artiehumphreys:
New Trigram search logic for courses, including indexes over title and "<mnemonic> <number>". Deprecated regex search that can be costly computationally.


tcf_website/views/search.py
```diff
@@ -131,47 +120,34 @@ def fetch_instructors(query):
...
-
-    # search query of form "<MNEMONIC><NUMBER>"
-    if number != "":
-        TITLE_WEIGHT = 0
-        MNEMONIC_WEIGHT = 1
-    # otherwise, "title" is entire query
-    else:
-        NUMBER_WEIGHT = 0
 
+    similarity_threshold = 0.3
```
barrett-ruth, line 126: magic

barrett-ruth: (explain)

barrett-ruth: also, how does this perform on searches for course titles? (i.e. not the mnemonics)

pls do just like 5 minutes of f*dging around with random searches n stuff before - once u say its fine we can merge

artiehumphreys: The 0.3 is for partial matches, in the case that a user searches a shorthand version of a class with a long name (ex: "data structures" when looking for "data structures and algorithms II"). Will do more testing now.

artiehumphreys: Changed similarity to 0.25 for a lower barrier for partial title searches. It seemed to work well with different searches.

barrett-ruth (approved):
nice

PR #970, opened 2024-11-25
UV

barrett-ruth:
Prefer uv to pip for CI/CD for performance reasons.

UV is used as a drop-in replacement for pylint, django, and GAE deployment

~30 second faster


artiehumphreys (approved):
Looks good. Thanks for doing this.

PR #989, opened 2025-01-29
Deprecate lfs, use gdown for downloading latest.sql

angadv12:
## GitHub Issues addressed

- This PR closes

## What I did
- Deprecated git-lfs for tracking latest.sql file.
- Added gdown to setup.sh script to check for gdown install and auto download latest.sql from tCF Google Drive using file ID in sharing link.
- Updated dev.md to include gdown as prereq prior to running one command curl setup.

## Screenshots

- Before
image
- After
image

## Testing

- Ran test fresh setup using `curl -fL 'https://raw.githubusercontent.com/thecourseforum/theCourseForum2/refs/heads/master/scripts/setup.sh' | sh -s -- <install_dir>`. Correctly pulled latest.sql from Google Drive and placed into db/latest.sql.

- Future: can research method of scanning google drive db files to download most recent one (can normalize method of naming files using dates and use that to scan files)


.gitattributes
```diff
@@ -4,4 +4,3 @@
 # e.g. `/bin/sh^M: bad interpreter: No such file or directory`
 # https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings
```
barrett-ruth, line 5: Just remove the entire file, p sure its just useless now

doc/dev.md
```diff
@@ -1,6 +1,6 @@
 # tCF Developer Info
 
-Ensure your system has [Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git), [git-lfs](https://git-lfs.com/) and [Docker](https://docs.docker.com/install/) installed.
+Ensure your system has [Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git), [git-lfs](https://git-lfs.com/), [Docker](https://docs.docker.com/install/), and [gdown](https://github.com/wkentaro/gdown) installed.
```
barrett-ruth, line 3: Thanks for adding this!

barrett-ruth (changes requested):

requirements.txt
```diff
@@ -8,6 +8,7 @@ django-filter~=23.5
 django-heroku~=0.3.1
 django-stubs~=4.2.7
 djangorestframework~=3.14.0
+gdown~=5.2.0
```
barrett-ruth, line 11: GDown is only necessary for the setup script, not for the docker continaer. I think this should be removed - what do you think?

angadv12: You're right I didn't even think of that, I'll remove.

PR #998, opened 2025-02-04
MVP for Enhanced Search

Jay-Lalwani:
## Overview

- This PR is to thoroughly test and eventually merge a production-ready enhanced search filtering functionality

## Testing

- Functionality Testing:

- UI/UX Testing:

-


tcf_website/views/browse.py
```diff
@@ -82,7 +82,7 @@ def department(request, dept_id: int, course_recency=None):
         request,
         "department/department.html",
         {
-            "subdepartments": dept.subdepartment_set.all(),
+            # "subdepartments": dept.subdepartment_set.all(),
```
ajnye, line 85: Remove this completely if not needed.

tcf_website/views/search.py
```diff
@@ -135,9 +155,19 @@ def normalize_search_query(q: str) -> str:
         .filter(Q(number__isnull=True) | Q(number__regex=r"^\d{4}$"))
         # filters out classes that haven't been taught since Fall 2020
         .exclude(semester_last_taught_id__lt=48)
-        .order_by("-max_similarity")[:10]
+        .order_by("-max_similarity")
     )
 
+    # Apply filters
```
ajnye, line 161: Seems like time filters are not applied when you query something with a search, like searching "CS 2100", is that correct? If so, it should be fixed.

tcf_website/views/search.py
```diff
@@ -154,3 +184,50 @@ def normalize_search_query(q: str) -> str:
     ]
 
     return courses
+
+
+def filter_courses(filters):
```
ajnye, line 189: Could also consider combining these methods somehow, as having two filter blocks of code to maintain seems redundant. Perhaps extract the filter/query DB results to separate functions that are called on the Course.objects?

tcf_website/views/search.py
```diff
@@ -154,3 +184,50 @@ def normalize_search_query(q: str) -> str:
...
+    if filters.get("disciplines"):
+        results = results.filter(disciplines__name__in=filters.get("disciplines"))
+
+    if filters.get("subdepartments"):
+        results = results.filter(subdepartment__mnemonic__in=filters.get("subdepartments"))
+
+    if filters.get("instructors"):
+        results = results.filter(section__instructors__id__in=filters.get("instructors"))
+
+    # Apply time filters
```
ajnye, line 209: Need to add this to the query function

tcf_website/views/search.py
ajnye: You could make the fetch_courses(query) handle the queries for both a valid search query and empty "" queries, with an if statement for the results, with a call to a general "filters" function that applies the 4 filters currently.

ajnye (changes requested):
Left some general comments for right now that should be changed

tcf_website/templates/search/searchbar.html
```diff
@@ -1,13 +1,536 @@
...
+      <button 
+        class="btn btn-outline-secondary border border-left-0" 
+        id="search-button" 
+        aria-label="Search" 
+        type="submit"
+      >
+        <i class="fa fa-search" aria-hidden="true"></i>
       </button>
     </div>
+    <div class="dropdown-wrapper">
```
ajnye, line 24: Separate out to another html to { % include ___ with ___ % } to separate functionality of dropdown and searchbar.

PR #1008, opened 2025-02-09
Default sorting option for course - instructor reviews

artiehumphreys:
## Completed:
- Created "Default" sorting method for reviews so that only the first ten reviews are fetched when a course instructor is loaded.
- This mode is not selectable and fetches the earliest reviews.


barrett-ruth (approved):
xd

PR #1038, opened 2025-02-28
Search optimize

Jay-Lalwani:
## GitHub Issues addressed

- This PR closes

## What I did

- cleaned up repeated db hits
- added basic cache

## Screenshots

- Before
- After

## Testing

-


tcf_website/views/search.py
```diff
@@ -8,24 +8,35 @@
...
-            }
-        grouped_courses[course_dept]["courses"].append(course)
+            subdept = subdepts.get(course_dept)
+            if subdept:
+                grouped_courses[course_dept] = {
+                    "subdept_name": subdept.name,
+                    "dept_id": subdept.department_id,
+                    "courses": [],
+                }
+        if course_dept in grouped_courses:
```
barrett-ruth, line 38: this if is unecessary due to control flow

tcf_website/views/search.py
```diff
@@ -8,24 +8,35 @@
...
         if course_dept not in grouped_courses:
-            subdept = Subdepartment.objects.filter(mnemonic=course_dept).first()
-            # should only ever have one returned with that mnemonic
-            grouped_courses[course_dept] = {
-                "subdept_name": subdept.name,
-                "dept_id": subdept.department_id,
-                "courses": [],
-            }
-        grouped_courses[course_dept]["courses"].append(course)
+            subdept = subdepts.get(course_dept)
```
barrett-ruth, line 31: use the walrus operator

tcf_website/views/search.py
```diff
@@ -208,22 +225,29 @@ def filter_courses(filters):
...
 
+
 def apply_filters(results, filters):
-    """Apply filters to course queryset."""
+    """Apply filters to course queryset using Q objects for more readable code."""
+    filter_conditions = Q()
+
     if filters.get("disciplines"):
-        results = results.filter(disciplines__name__in=filters.get("disciplines"))
+        filter_conditions &= Q(disciplines__name__in=filters.get("disciplines"))
```
barrett-ruth, line 234: nice use of &, didn't know that

tcf_website/views/search.py
```diff
@@ -208,22 +225,29 @@ def filter_courses(filters):
...
+    if filter_conditions:
+        results = results.filter(filter_conditions)
 
+    # Handle time filters
     weekdays = [day for day in filters.get("weekdays", []) if day]
     from_time = filters.get("from_time")
     to_time = filters.get("to_time")
 
-    if len(weekdays) != 5 and len(weekdays) != 0 or from_time or to_time:
+    if (len(weekdays) != 5 and len(weekdays) != 0) or from_time or to_time:
```
barrett-ruth, line 250: remove superfluous parens (opinion)

barrett-ruth: prefer `weekdays` over `len(weekdays) != 0)`, which should be `len(weekdays > 0)` regardless

tcf_website/views/search.py
```diff
@@ -208,22 +225,29 @@ def filter_courses(filters):
 
     return courses
 
+
 def apply_filters(results, filters):
```
barrett-ruth, line 229: Is it more efficient to applies all three of these filters in a joined condition at once, rather than iteratively?

Speaking in terms of django under the hood.

tcf_website/views/search.py
```diff
@@ -135,30 +156,26 @@ def normalize_search_query(q: str) -> str:
...
-            title_similarity=TrigramSimilarity("title", search_query),
-        )
-        # round results to two decimal places
-        .annotate(
+            mnemonic=F("subdepartment__mnemonic"),
             max_similarity=Round(
                 Greatest(
-                    F("mnemonic_similarity"),
-                    F("title_similarity"),
+                    TrigramSimilarity("combined_mnemonic_number", search_query),
```
barrett-ruth, line 162: Why is this being changed from before? @artiehumphreys note this change.
Just looking for some explanation as we spent a bit of time making sure this was right.

tcf_website/views/search.py
```diff
@@ -50,14 +61,24 @@ def search(request):
     # Save filters to session
     request.session["search_filters"] = filters
 
-    if query:
+    # Create a cache key from the query and filters
+    cache_key = f"search_{query}_{hash(frozenset(str(filters.items())))}"
+    cached_results = cache.get(cache_key)
```
barrett-ruth, line 66: This just makes me nervous. Is there really no way to cache? Is below not what we're looking for?

https://docs.djangoproject.com/en/5.1/topics/cache/

```python
from django.views.decorators.cache import cache_page

@cache_page(60 * 15)
def my_view(request): ...
```

barrett-ruth (changes requested):

tcf_website/views/search.py
Jay-Lalwani: it's the same just not storing intermediate results, but i reverted because it doesn't make much speed difference (if at all) and it's easier to debug as it was originally.

PR #1039, opened 2025-03-01
Create prune-gh-branches.sh

alex-yung-github:
created a script to prune branches of a certain age (based on most recent edit)


hari-gajjala (approved):
lgtm
nice commenting

barrett-ruth (changes requested):
Hi Alex, this is fire. Just some advice for scripting: 

- rather than stating in a comment that certain executing must exist, you can verify this the the case with something like `type` or `command` (just as chatgpt).
- same thing with logging in - why not just run `gh auth login` inside the script in the first place?

PR #1107, opened 2025-04-08
fix: server 500 on partial credits for schedules

kmd2zjw:
## What I did

- change the schedule model to allow partial credits

## Testing

- ran locally on site with CGBM classes in schedule, running through edit and add operations to the schedule


ajnye (approved):
LGTM

PR #1109, opened 2025-04-10
Moderation

blairezh:
## GitHub Issues addressed

- This PR closes issue of inappropriate reviews on tcf.

## What we did

- Created command script using detoxify to evaluate reviews 
- When evaluated, reviews are assigned a rating 1-100 for toxicity and the most relevant descriptive category of toxicity
- Removed display of reviews with a rating of 74 or higher

## Screenshots

- Before (review text covered up)
image

- After (review is now gone)

image

## Testing

- Reviewed ratings + checked an example instructor course page to see if review was hidden
- If you want to look at which reviews are being filtered out, run the command with the --results flag

- UI adjustments needed (bug fix!)
- This is only a temporary solution as the model doesn't distinguish between negativity towards the course, negativity towards the professor/TAs, and negativity towards the reviewer, as well as positive vs negative cursing. Also, this only retroactively filters out reviews in batches.
- For the future, we should increase accuracy in detecting harmful reviews vs helpful reviews either with a different model or fine-tuning detoxify (toxic-bert) and implement real-time moderation.


barrett-ruth (approved):
ty

PR #1112, opened 2025-04-11
Added Search pagination 

YuDavidCao:
## GitHub Issues addressed

- This PR closes

## What I did

- Added search pagination

## Screenshots

- Before
Screenshot 2025-04-11 at 6 24 56 PM

- After
Screenshot 2025-04-11 at 6 32 56 PM

## Testing

- checked if page number is correct by clicking the different buttons

-


barrett-ruth:
@ajnye @Jay-Lalwani

angadv12:
changed single / double arrow button functionality to align w review and dept pagination style

barrett-ruth:
@angadv12 @YuDavidCao I'm hesitant to perform this due to performance reasons. I mean, there's 478 results...  (nobody is clicking that far) not to mention that the aws bill is through the roof? What do y'all think about either approach

1. Limiting the total number of results to just "n" pages
2. A "lazy" search. If I click next and I'm on page `n` showing `k` search results per query, you then re-run the search and indexthe search results by `[(n+1)*k:(n+2)*k]`

YuDavidCao:
I did a bit of performance testing and I don't see a significant increase on the search results impacted by the pagination. Here's a video on how I did the testing (all page refresh are cmd + shift + R instead of cmd + R)

https://github.com/user-attachments/assets/03b8d6f7-93e8-4eba-b9d8-312c11f2cfb4

Here's the performance metrics of implementing pagination:
image

Here's the performance metrics of using plain limiting to 15 search results:
image

The top blue one is cpu, green one is memory, purple one is disk read + write, and the bottom one is network traffic. I would say that there's probably like 10% of cpu increase using pagination than limiting to 15 search results but I don't think its statiscally significant enough. 

For the approach 1, if I am thinking of implementing it I will still need to use the paginator and all the changes would be made on the templates side using a min function probably which doesn't impact performance much.

For the approach 2, we can make that but it feels like making a simple paginator because if I think about it I would also track a index (or "n" in your description) and multiply it by "k" and use an array slicing on the queryset. The only parts I would think that would outperform django paginator is getting the total number of results and the complexity of the entire package of django paginator. But the problem is that we need to handle the "next page is empty" situation which I am not exactly sure how that can be done without knowing the total number of results. 

It's all my judgement though feel free to inspect the video more.

Wdy think?

PR #1151, opened 2025-09-28
Optimize page size by removing search bar filters

gyoge0:
## GitHub Issues addressed

- N/A

## What I did

- Removed filtering by subject and discipline in the search bar

## Screenshots

- Before
Screenshot 2025-09-28 at 3 46 59 PM

- After
Screenshot 2025-09-28 at 3 47 07 PM

## Testing

- Checkout this commit and run docker compose up

- Priority for this PR is to remove the feature to save on page size - 50kb vs 250kb sent to users matters a lot, and should slash our data transfer out costs. 
- Adding back this feature should be saved for later, and should be done in a way that doesn't transfer this much data.
- We should pay attention to if people are even using this or not.

## Summary by CodeRabbit

- Refactor
  - Removed “Subject” and “Discipline” filters from the search filter panel, simplifying the filtering experience while keeping other filters (day/time, availability, GPA, etc.) unchanged.
- Style
  - Cleaned up related styles by removing list and custom scrollbar styling associated with the removed filters, resulting in a leaner, more consistent UI.

### DIFF
--- a/tcf_website/static/search/searchbar.css
+++ b/tcf_website/static/search/searchbar.css
@@ -202,34 +202,6 @@
   pointer-events: none;
 }
 
-.form-check-subjects,
-.form-check-disciplines {
-  margin-top: 1.2% !important;
-  align-self: start !important;
-}
-
-.subject-list,
-.discipline-list {
-  max-height: 150px;
-  overflow-y: auto;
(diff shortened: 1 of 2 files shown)


ldkohler:
@Jay-Lalwani deprecate advanced search to save on ELB?

PR #1246, opened 2026-04-13
Enhanced Search Results

artiehumphreys:
## Problem:
The search results are often polluted with irrelevant courses that either haven't been taught in a while (e.g. https://thecourseforum.com/search/?mode=courses&q=computer) or aren't popular among the students. This applies particularly for broader searches, such as those like "artificial intelligence" where the plethora of special topics courses related to the matter aren't displayed in the search results and instead there are many graduate law classes that only a small fraction of users will ever click.

## What this PR solves:
- Weighs search results on trigram similarity (as before) and also recency and number of reviews
- Adjustable weights 

## Summary by CodeRabbit

* **New Features**
  * Search results now use a weighted ranking that combines relevance, review popularity (normalized, with an upper cap) and recency of offering to surface more relevant courses.
  * Results still apply a minimum relevance threshold to filter out low-quality matches.
  * Improved ordering produces more useful top results reflecting student interest and recent offerings.


tcf_website/views/catalog/search.py
```diff
@@ -46,23 +50,51 @@ def normalize_search_query(raw: str) -> str:
...
+            ),
+            recency_score=Greatest(
+                Value(0.0),
+                Value(1.0)
+                - (Value(latest_sem_number) - F("semester_last_taught__number"))
+                / Value(
+                    50.0
+                ),  # anything last taught 5 years ago or later has a score of 0
+                output_field=FloatField(),
+            ),
```
gyoge0, line 85: are we able to suppress the ruff line length here?

gyoge0 (dismissed):

Jay-Lalwani:
How does this affect performance?

review_count=Count("review", distinct=True) looks very expensive.

artiehumphreys:
> How does this affect performance?
> 
> review_count=Count("review", distinct=True) looks very expensive.

The affect on performance is minimal as each search already fetches review data (`LEFT INNER JOIN` in the picture below) in each call to `Course.with_stats()` which is called by `browseable_course_queryset`. This means that fetching the number of reviews is mostly trivial. Below are results of cold searches on both `dev` and enhanced-search-results` for a simple query

# `enhanced-search-results`:

Screenshot 2026-04-13 at 12 37 27 PM

# `dev`

Screenshot 2026-04-13 at 12 36 55 PM

-----
Naturally, there is a little more overhead but its relatively small compared to the overhead of the Trigram on every course.

PR #1254, opened 2026-04-26
fix: notify user when resetting password with unregistered email (#1251)

### LINKED ISSUE #1251: Bug: User not notified if resetting password with unregistered account
## Describe the bug

If a user attempts to sign in via 2FA or reset the password for an account with an email that isn't in the system, the user is not notified that there isn't an account that exists with that email. Instead, no email is ever sent to the user and they have no idea that there is no account associated with that email. It's a subtle bug but may be troublesome for some users.

## Expected behavior

A notification stating that there is no account associated with the email provided, prompting the user to create an account.

CrepuscularIRIS:
Fixes #1251

## What's happening now

When a user visits the Cognito-hosted forgot-password flow and enters an email that isn't registered, Cognito silently does nothing (its default "prevent user existence errors" mode). The user never gets any email and has no idea why.

## What this PR does

Adds a custom `/forgot-password/` page that intercepts the request before it reaches Cognito:

1. The user enters their email on the new Django page.
2. The view checks whether that email matches any account in the local database (case-insensitive, so mixed-case emails stored from Cognito JWT claims are handled correctly).
3. If no match is found, the user immediately sees: *"No account is associated with that email address. Sign in with your UVA email to create an account."*
4. If a match is found, the view calls `cognito-idp.forgot_password` via boto3 directly, which triggers the reset email. A `UserNotFoundException` from Cognito is treated as a second safety net showing the same message.
5. On success, the user is redirected to the login page with a "check your inbox" confirmation.

A "Forgot your password?" link is also added to the existing login modal.

## A note on design tradeoffs

I might be wrong about this, but I think for a university forum where all accounts are UVA email-based, explicitly telling someone "no account found" is more useful than the vague silence they get today. If the maintainers prefer a generic "if your account exists, we sent a link" message for security reasons, I'm happy to adjust — just let me know and I'll update the success/error paths.

Rate-limiting this endpoint (e.g. with django-ratelimit or a CAPTCHA) would be a sensible follow-up but felt out of scope for this fix.

## Files changed

- `tcf_website/views/auth/__init__.py` — new `forgot_password` view + `_cognito_secret_hash` helper
- `tcf_website/views/__init__.py` — re-exports `forgot_password`
- `tcf_website/urls.py` — `/forgot-password/` URL
- `tcf_website/templates/site/auth/forgot_password.html` — new template
- `tcf_website/templates/site/common/modals/_login.html` — "Forgot your password?" link
- `tcf_website/tests/test_auth.py` — 7 unit tests (GET, empty email, unknown email, mixed-case email, Cognito success, `UserNotFoundException`, generic Cognito error)

---

*This contribution was written with AI assistance (Claude). All logic has been reviewed for correctness. If this direction doesn't fit the project's approach, no problem — happy to close or rework.*

### DIFF
--- a/tcf_website/templates/site/auth/forgot_password.html
+++ b/tcf_website/templates/site/auth/forgot_password.html
@@ -1,0 +1,38 @@
+{% extends "site/common/base.html" %}
+{% load static %}
+{% block title %}
+    Forgot Password – theCourseForum
+{% endblock title %}
+{% block content %}
+    <div style="max-width: 440px; margin: 4rem auto; padding: 0 1rem;">
+        <h1 class="mb-4">Reset your password</h1>
+        {% if messages %}
+            {% for message in messages %}
+                <div class="alert {{ message.tags }} mb-4" role="alert">{{ message }}</div>
+            {% endfor %}
+        {% endif %}
+        <p class="text-muted mb-4">
+            Enter the email address linked to your theCourseForum account and we will send you a
+            password reset link.
+        </p>
+        <form method="post">
+            {% csrf_token %}
+            <div class="mb-4">
+                <label for="email" class="form-label">Email address</label>
+                <input type="email"
+                       id="email"
+                       name="email"
+                       class="form-control mt-1"
+                       value="{{ email|default:'' }}"
+                       placeholder="computing_id@virginia.edu"
+                       required
+                       autofocus>
+            </div>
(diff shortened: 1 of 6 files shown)


CrepuscularIRIS:
Closing this PR — no maintainer activity in 2+ weeks. Happy to reopen or resubmit if you'd like to revisit. Thanks for the project!

PR #1259, opened 2026-07-04
1253 display min char limit for reviews

### LINKED ISSUE #1253: Style: Display minimum character limit for reviews
When writing reviews for a course, instead of displaying the character count in red font when there aren't enough characters, there should be some display that shows the number of required characters for a review.

Image

Adrian-E-V:
## GitHub Issues addressed

- This PR closes issue 1253. "Style: Display minimum character limit for reviews"

## What I did

- Reviews show the number of required characters, counting down from 200 to 1. Once reaching 200 characters, the character count will count up as normal.
- Changed hard coded value 200 for textarea.minLength
- review.html is the only file modified

## Screenshots

- Before
image

- After
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the live review text counter to use the field’s configured minimum length instead of a fixed threshold.
  * When the input is below the minimum, the counter shows `current/min required characters` and uses an alert (danger) styling.
  * Once the minimum is met or exceeded, it switches back to displaying the current character count in a muted style.

### DIFF
--- a/tcf_website/templates/site/review/review.html
+++ b/tcf_website/templates/site/review/review.html
@@ -184,7 +184,7 @@
       const characters = text.length;
       countDisplay.textContent = characters + ' characters';
       if (characters < 200) {
-          countDisplay.style.color = 'var(--danger)';
+          countDisplay.style.color = 'Black';
       } else {
           countDisplay.style.color = 'var(--fg-muted)';
       }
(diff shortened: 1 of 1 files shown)


jackrhoa:
Can we show the target also, not just the remaining count?

Adrian-E-V:
yup, how is this @jackrhoa ?
image

It will count up normally after reaching 200

jackrhoa:
> yup, how is this @jackrhoa ?
> 
> image
> 
> 
> 
> It will count up normally after reaching 200

Looks good. Going to merge

jackrhoa (approved):

PR #1272, opened 2026-08-24
fix(actions): fix github actions for node 24

YuDavidCao:
## GitHub Issues addressed

N/A

## What I did

- Github is deprecating Node 20 in their actions by September 16th 2026, those actions that are being updated still is based on node 20. Those actions are currently being forced to run at node 24. And I figured that those are rather harmless to upgrade to the newest version instead of stamping them with a `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true`

Screenshot 2026-08-24 at 12 01 53 PM

This is the official github blog post on node 20 deprecation: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

## Screenshots

- Before
- After

## Testing

-

## Summary by CodeRabbit

* **Chores**
  * Updated automated build, test, and deployment workflows to use newer, more reliable tooling.
  * Improved compatibility across continuous integration and AWS deployment processes.
  * Refreshed the testing environment, including runtime and database versions, to support current platform requirements.
  * Standardized continuous integration configuration for more consistent validation across development workflows.


YuDavidCao:
non of those coderabbit comments are specifically to this commit. They are choices made earlier, worth looking separately but not a blocker for this PR.

gyoge0 (commented):
While we are making these changes, can we pin the new versions to commit hashes instead of just the `@vX`? Also, did we test these actions to see if they are pulling in the right versions? [CI ran on this PR](https://github.com/thecourseforum/theCourseForum2/actions/runs/32749057207/job/97501387831) but still looks like it's using Node 20 and not 24. Maybe we could use [act](https://github.com/nektos/act) to test it locally to confirm.

If we can pin new versions to commit and verify the right Node version is running, we can merge.

YuDavidCao:
@gyoge0 good call on the eslint ci step, forgot to change that from 20 to 24, just changed. To your question:

1. can we pin the new versions to commit hashes instead of just the @ vX

I assume you meant the setup-uv, from their official docs, pinning the hash commit is the way to go so I went for it: https://github.com/astral-sh/setup-uv, I also cannot find any reference to pinning it with a version so

2. Also, did we test these actions to see if they are pulling in the right versions?

I tested ci.yml, worked just fine. Not going to test aws.yml, it needs aws credential, I am not going to run it as I think it might mess things up. But generally I don't think github action update would change anything, it should be a very low risk change

3. CI ran on this PR Node 20 and not 24.
I updated that step to use node-version 24, good catch

jackrhoa (commented):
Looks good, will wait for Yogesh to take a look before merging

gyoge0:
> I assume you meant the setup-uv, from their official docs, pinning the hash commit is the way to go so I went for it: https://github.com/astral-sh/setup-uv, I also cannot find any reference to pinning it with a version so

Yes, can we do this for all of the actions and not just uv? After that we can merge.

YuDavidCao:
I think setup-uv is the only one that's doing this, the other github and aws ones use versions.

@gyoge0

jackrhoa:
> Yes, can we do this for all of the actions and not just uv? After that we can merge.

Addressed in #1288

PR #1273, opened 2026-08-25
feat: explicit TCF_ENV runtime mode

gyoge0:
Introduce a single explicit environment variable (TCF_ENV=local|ci|prod)
that declares which mode the process runs in, replacing inference from
side effects like GITHUB_ACTIONS or whether debug_toolbar happens to be
installed.

- base.py: parses and validates TCF_ENV at boot; bad values refuse to start
- dev.py: DEBUG and debug toolbar keyed off ENVIRONMENT == "local";
  opt-in prod-parity backends behind REDIS_URL / MINIO_ENDPOINT /
  STATIC_S3 (all default off, so behavior is unchanged unless enabled)
- prod.py: refuses to boot unless TCF_ENV=prod
- ci.yml sets TCF_ENV=ci; iac/ecs.tf sets TCF_ENV=prod

## Summary by CodeRabbit

* **New Features**
  * Added environment-specific configuration for local, CI, and production deployments.
  * Enabled optional Redis-backed caching and database sessions when configured.
  * Added optional MinIO media storage and manifest-based static file storage.
  * Debug tools are now available only in local development environments.

* **Bug Fixes**
  * Applications now fail clearly when an unsupported environment is configured or production settings are used incorrectly.

* **Documentation**
  * Documented supported environment modes and updated local CI setup instructions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
 env:
   PYTHON_TARGET: "3.12"
   DJANGO_SETTINGS_MODULE: tcf_core.settings.dev
+  TCF_ENV: ci
   SECRET_KEY: ci-secret-key-not-for-production
   # CI database (GitHub Actions postgres service)
   DB_NAME: tcf_db
@@ -80,7 +81,7 @@ jobs:
       code-coverage: ${{ steps.coverage.outputs.percentage }}
     services:
       postgres:
-        image: postgres:15.4
+        image: postgres:18.1
         env:
           POSTGRES_USER: ${{ env.DB_USER }}
           POSTGRES_PASSWORD: ${{ env.DB_PASSWORD }}
--- a/iac/ecs.tf
+++ b/iac/ecs.tf
@@ -37,6 +37,10 @@ container_definitions = jsonencode([
           value = "tcf_core.settings.prod"
         },
         {
+          name  = "TCF_ENV"
+          value = "prod"
+        },
+        {
           name  = "AWS_STORAGE_BUCKET_NAME"
           value = aws_s3_bucket.static.id
         },
--- a/tcf_core/settings/base.py
+++ b/tcf_core/settings/base.py
@@ -4,18 +4,23 @@ import os
 
 import environ
 from django.contrib.messages import constants as messages
+from django.core.exceptions import ImproperlyConfigured
 from django.urls import reverse_lazy
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
 # Django-environ library imports .env settings
-env = environ.Env(
-    # set casting, default value
-    DEBUG=(bool, False),
-)
+env = environ.Env()
 env_file = os.path.join(BASE_DIR, ".env")
 environ.Env.read_env(env_file)
+
+# Runtime mode, set by the launcher: local (default) | ci | prod.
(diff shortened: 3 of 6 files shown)


iac/ecs.tf
```diff
@@ -36,6 +36,10 @@ container_definitions = jsonencode([
           name  = "DJANGO_SETTINGS_MODULE"
           value = "tcf_core.settings.prod"
         },
+        {
+          name  = "TCF_ENV"
+          value = "prod"
```
jackrhoa, line 41: We'll also need to set this env var in the AWS Console since we're not deploying with IaC yet.

tcf_core/settings/dev.py
```diff
@@ -1,11 +1,9 @@
...
+"""Django settings for local development and CI."""
 
-GitHub Actions sets GITHUB_ACTIONS=true on the runner; local shells do not.
-"""
+import importlib.util
 
 from .base import *
-
-_ci = os.environ.get("GITHUB_ACTIONS") == "true"
+from .base import ENVIRONMENT  # noqa: F401
```
jackrhoa: ENVIRONMENT is already imported by `from .base import *` on line 5. No need to import again

tcf_core/settings/base.py
jackrhoa: This line does nothing now since both `dev.py` and `prod.py` set this value

tcf_core/settings/dev.py
```diff
@@ -23,13 +21,74 @@
...
+            "OPTIONS": {
+                "socket_connect_timeout": 5,
+                "socket_timeout": 5,
+                "retry_on_timeout": True,
+            },
+        }
+    }
+    SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
+    SESSION_CACHE_ALIAS = "default"
+    CACHALOT_ENABLED = True
```
jackrhoa, line 45: We should probably also set `CACHALOT_TIMEOUT`. The default is to never expire

tcf_core/settings/dev.py
```diff
@@ -1,11 +1,9 @@
-"""Django settings for local development and CI.
+"""Django settings for local development and CI."""
 
-GitHub Actions sets GITHUB_ACTIONS=true on the runner; local shells do not.
-"""
```
jackrhoa, line 4: `doc/dev.md` needs to be updated as well to reflect the changes

.github/workflows/ci.yml
```diff
@@ -80,7 +81,7 @@ jobs:
       code-coverage: ${{ steps.coverage.outputs.percentage }}
     services:
       postgres:
-        image: postgres:15.4
+        image: postgres:18.1
```
jackrhoa, line 84: Our prod db is currently running pg 17.7.

iac/ecs.tf
gyoge0: I set this in the ECS task def already for the legacy account

tcf_core/settings/dev.py
gyoge0: Removing cachalot in the future

.github/workflows/ci.yml
gyoge0: IAC and tcf3 are both on 18.1 I believe, might hit some issues with this come dev setup but I think we can work it out then.

tcf_core/settings/dev.py
gyoge0: done

tcf_core/settings/base.py
gyoge0: fixed

jackrhoa (approved):
looks good to me

PR #1286, opened 2026-09-07
feat: fetch grades using API

jackrhoa:
Rewrote grade scraping script into a management command that uses the grade distribution WebSocket API. Takes < 1 min per semester

## Summary by CodeRabbit

- **New Features**
  - Added a management command for fetching semester grade data.
  - Added `--min-rows` and `--output` options for controlling validation and output location.
  - The command uses instructor names from semester data when available and falls back to registrar names with a warning.

- **Documentation**
  - Updated grade-data instructions with the new command invocation and options.
  - Updated production guidance to recommend running the command through `ecs-run-command.sh`.


gyoge0 (changes requested):
Will approve after the selenium/websocket issue is fixed

 [P1] Missing runtime dependency — fetch_grades.py:31 imports websocket, but websocket-client is only installed  
   transitively through the dev-only Selenium dependency. Production Docker builds use uv sync --no-dev, so the    
   command fails with ModuleNotFoundError.

gyoge0 (approved):

Issue #1292, opened 2026-09-10
Fix grades not attributed to correct instructor


jackrhoa:
Grades are not attributed to the correct instructor because grades are attributed to the instructor listed for the semester data. In many instances, semester data is loaded before instructors are finalized and never updated.

Instructor names in grade data are legal names, whereas semester data (from SIS) uses preferred names that students are more familiar with. 

One potential solution would be to re-fetch the SIS data for previous semesters, taking special care to ensure no reviews are lost where they shouldn't be.

PR #1294, opened 2026-09-12
fixing advanced search date filter by normalizing date to lower case due to inconsistency in csv files

YuDavidCao:
## What I did

- I noticed that the advance search dropdown's date filter is currently not working for Fall 2026 courses when I was browsing. It turned out that the date representation in the csv has different cases (some upper, some lower), and the Fall 2026 is the lower case one, which I is the problem. This PR normalized the date filter to lower case.

## Testing

- I loaded courses for Fall2026, did not work locally, added the fix, worked for me locally.

## Note:

This would require a semester reload on prod.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of meeting-day data with varying capitalization.
  * Preserved the original casing of imported meeting-day values while continuing to display unavailable days as “TBA”.
  * Improved parsing of weekday abbreviations in semester section schedules.

### DIFF
--- a/tcf_website/management/commands/fetch_data.py
+++ b/tcf_website/management/commands/fetch_data.py
@@ -151,7 +151,7 @@ def compile_course_data(course_number, sem_code):
         if not m:
             return ""
         meets = m["meets"]
-        return (meets if meets != "-" else "TBA").lower()
+        return meets if meets != "-" else "TBA"
 
     def _meeting_room(m) -> str:
         if not m:
--- a/tcf_website/management/commands/load_semester.py
+++ b/tcf_website/management/commands/load_semester.py
@@ -397,14 +397,17 @@ class Command(BaseCommand):
                 try:
                     days_part, time_part = time_block.strip().split(" ", 1)
                     start_time, end_time = time_part.split(" - ")
+                    
+                    # CSV casing varies, normalize to lowercase.
+                    days_part = days_part.lower()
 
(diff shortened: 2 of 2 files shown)


jackrhoa:
This issue comes from `fetch_data.py` making day of week lowercase. The non-lowercase version (Mo, Tu, We, etc. ) would be preferred when displaying data on the site. Could you update this PR to address that root cause as well?

YuDavidCao:
> This issue comes from `fetch_data.py` making day of week lowercase. The non-lowercase version (Mo, Tu, We, etc. ) would be preferred when displaying data on the site. Could you update this PR to address that root cause as well?

@jackrhoa addressed

Prod still needs re-load to make it work though

gyoge0 (commented):
How will this work with data already stored in the DB? Do we need a migration for this? I can see old data within the DB still having "Fr" and not having the `.lower()` call will make it not match.

jackrhoa:
> How will this work with data already stored in the DB? Do we need a migration for this? I can see old data within the DB still having "Fr" and not having the `.lower()` call will make it not match.

Data already stored in the DB as "Fr" will be fine because it has already been properly mapped with `load_semester.py`. We could run a migration command to map the unmapped courses, but given that we will need to reload courses to address #1292 and the add deadline has passed, I think it makes sense to just make these changes to fix this issue for future semesters.

PR #1295, opened 2026-09-12
added pre commit hook for linting

YuDavidCao:
## What I did

- I got a linting error on one of my earlier PR, and I remember getting this a lot. So, to prevent this, I added a pre commit hook to run uv run ruff check --fix on the staged python files 

## Testing

- When I commited this commit, I deliberatedly added a tab on an empty line in a py file, it caught it and auto fixed it.


gyoge0 (changes requested):
Can we use [prek](https://prek.j178.dev)?

YuDavidCao:
@gyoge0 changed to prek

.gitignore
```diff
@@ -38,6 +38,8 @@ venv
 .venv
 # Ruff
 .ruff_cache/
+# prek (generated by `uv run prek install`)
+.husky/
```
gyoge0: Is this correct

YuDavidCao:
Screenshot 2026-09-12 at 6 50 09 PM

@gyoge0 I didn't know this but I guess using husky as dir is the naming convention for pre commit hook

gyoge0:
> @gyoge0 I didn't know this but I guess using husky as dir is the naming convention for pre commit hook

I think it's because prek picked up on you having husky already configured? This isn't the default behavior for a fresh repo

.gitignore
gyoge0:
```suggestion
```

gyoge0 (changes requested):

YuDavidCao:
@gyoge0 your right, it's picking up the core.hooksPath from my previous husky setting, did it again and it's now under .git/hooks/pre-commit which is already in the gitignore, removed the line for gitignore in the husky dir.

gyoge0 (approved):
🔥

jackrhoa (approved):

